PR #2, opened 2016-01-09
Use Gemfile for bundler

harding:
Create Gemfile, updated README.md and .travis.yml to use it with
bundler. This provides the following features:

- Travis will automatically pull the dependencies from the Gemfile

- Travis will cache the dependencies, speeding up CI testing

- Simplifies instructions if we add new dependencies

- Gemfile.lock makes sure everyone has exactly the same version of the
  dependencies, helping to stop some "works for me"/"doesn't work
  for me" bugs

- Devs who use rvm will automatically have the correct version of `ruby`
  selected for them

Also removed ruby 2.2.3 from the .travis.yml since it causes the build
to run twice for each ruby version.  Since we're not not doing app
development, we don't usually need to worry about forward compatability
at the minor-point-release level.

Also also corrected indent for env variable to get rid of a Travis error.


.travis.yml
```diff
@@ -1,20 +1,19 @@
...
 
-# branch whitelist
-branches:
-  only:
-  - gh-pages     # test the gh-pages branch
-  - /pages-(.*)/ # test every branch which starts with "pages-"
+## Disable external link checking to prevent spurious failures because
+## of other people's downtime. This can also avoid wasting their
+## bandwidth.
+script: bundle exec jekyll build && bundle exec htmlproof --disable-external ./_site
```
btcdrak: Do we really want to not check external links? I thought it was a good check to catch links that go stale months or years later when no-one is looking.

harding: I think it's a good manual check that can be run time-to-time.  But as an automated check it has two problems:

1. It causes spurious CI failures when a server is temporarily unreachable.
2. It sends the message that people should be polluting their PRs with unrelated link changes in order to fix CI failures.

PR #6, opened 2016-01-11
Use 1.5em line height

jonasschnelli:
Most pages noways have a line-height of >1em.
Using a line height of 1.5em makes things more readable.

Before:
bildschirmfoto 2016-01-11 um 15 31 54

After:
bildschirmfoto 2016-01-11 um 15 31 29

### DIFF
--- a/_sass/page.scss
+++ b/_sass/page.scss
@@ -221,6 +221,7 @@ $button-size: 1.5rem;
 	counter-reset: captions;
 	@include container;
 	@include clearfix;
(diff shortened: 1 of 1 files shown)


btcdrak:
Doesn't seem to have any effect on `/en/blog/` page.

jonasschnelli:
Fixed. Blog is also covered now.

btcdrak:
Tested ACK d9340cb

PR #11, opened 2016-01-12
Fix <li> syntax

MarcoFalke:
The list items appear smaller because they are no longer `h3`.

screenshot from 2016-01-12 18-25-05

[new vs old]


btcdrak:
Tested ACK fa3074f

PR #12, opened 2016-01-12
Format "Contribute"

MarcoFalke:
This has been a boring list. Now, there are two subsections and the senteces are connected.


btcdrak:
ACK

Issue #8, opened 2016-01-12
Improvements


MarcoFalke:
* ~~"Deutsche" in the translation dropdown should read "Deutsch" (#9)~~
* ~~Also shouldn't the English sources show up if no German translation is available? #13~~
* ~~Travis should run on pulls~~
* ~~The website should link to "https://github.com/bitcoin-core/website" (maybe in "About"?) (#10)~~
* The license is MIT, so the "© 2016 Bitcoin Core." text should link to the MIT license text. #15 
* ~~Design should look more "modern" and "trustworthy". (Maybe take it from https://bitcoin.org/en/bitcoin-core/ ?)~~
* Does the website comply with the EU cookie law (maybe something like http://www.imaginit.net/legal.cfm#q7)?

btcdrak:
> The website should link to "https://github.com/bitcoin-core/website" (maybe in "About"?)

It is linked in CONTRIBUTING.md

> The license is MIT, so the "© 2016 Bitcoin Core." text should link to the MIT license text.

Content cannot be MIT licensed, it would need to be some kind of CC. MIT covers code not design,

> Design should look more "modern" 

There's no problem with design overhauls or snazzing up the site but it should be distinct from bitcoin.org

MarcoFalke:
> Content cannot be MIT licensed

IANAL, but are you sure? Though, I agree using CC-BY or CC-BY-SA.

laanwj:
Afaik content and images can be MIT licensed (though it is rare for artists
to do so). If not we have a problem: The bitcoin core GUI images are
licensed as such.

laanwj:
Btw I agree that the english version of pages - esp blog items should be
shown if the translation is missing. E.g. content shouldn't be missing just
because a different lang is chosen.

btcdrak:
@laanwj "software" is different since it's a bundle with documentation "this software and associated documentation files". It would be more usual to license content with a specific content license like CC. The website source is MIT licensed already, however unlike software we have no obligation to do so, nor to make content freely copyable. In any case, if we want to license content we definitely should use a CC license.

Agree about non-translated blog items, it's on my to-do list, default to showing the English version if not translated.

laanwj:
Right - Whatever license it needs to be clearly stated, so that
contributors know what is expected

harding:
[...]

PR #13, opened 2016-01-12
Show English posts when no translations are available

btcdrak:
Display's English posts where no translation exists in blog listing

#8


MarcoFalke:
Concept ACK

PR #15, opened 2016-01-12
Update CONTRIBUTING.md with licensing information

btcdrak:
Code and documentation that make this website are licensed under MIT
Content published on the website suggested [CC-by-SA](http://creativecommons.org/licenses/by-sa/4.0/) which is the most permissive version.

#8

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,3 +54,12 @@ Translators should only translate the `title:` field, and change the language co
 Notice the remaining fields are left untranslated, however in Transifiex, you must copy these fields as translations (without actually translating them).
 
 Please note that [translations.yml](https://github.com/bitcoin-core/website/blob/gh-pages/_data/translations.yml) and [navigation.yml](https://github.com/bitcoin-core/website/blob/gh-pages/_data/navigation.yml) should be translated and submitted as normal pull-requests because they are not compatible with Transifex at this time.
+
+## License
+
+The code and documentation of this website are licensed under [MIT license][MIT] and all contributors agree to irrevocably license their contributions under the same license.
+
+Unless specified in the header of the file, the website content in `_posts/` of this website is licensed under the Creative Commons Attribution-ShareAlike 4.0 International License: [CC-BY-SA] and all contributors agree to irrevocably license their content under the same license.
+
+[MIT]: https://github.com/bitcoin-core/website/blob/gh-pages/LICENSE
+[CC-BY-SA]: http://creativecommons.org/licenses/by-sa/4.0/


CONTRIBUTING.md
```diff
@@ -27,3 +27,14 @@ and final decision is with the Bitcoin Core project lead.
...
 Translations should be made from the English source. The English source has a `version:` field in the Front Matter which should be updated if on the rare
 occasion the English text is updated. Translated text should remove `translated: false` from the Front Matter.
+
+## License
+
+The code and documentation of this website are licensed under [MIT license][MIT] and all contributors agree to license their
+contributions under the same license.
+
+The content of this website is licensed under the Creative Commons Attribution-ShareAlike 4.0 International License: [CC-By-SA]
+and all contributors agree to license their content under the same license.
```
MarcoFalke: You'd need to ask the previous contributors and translators if they agree.

MarcoFalke:
I'd prefer to just keep MIT.

CONTRIBUTING.md
```diff
@@ -27,3 +27,14 @@ and final decision is with the Bitcoin Core project lead.
...
+
+[MIT]: https://github.com/bitcoin-core/website/blob/gh-pages/LICENSE
+[CC-By-SA]: http://creativecommons.org/licenses/by-sa/4.0/
```
MarcoFalke: Maybe this should have it's LICENSE file as well?

CONTRIBUTING.md
luke-jr: I don't think we should commit to a uniform content license. We may very well wish to repost or quote others, and should be free to do so without negotiating licensing of it.

harding:
I prefer MIT-only for content being ported from Bitcoin.org.  @saivann and I are the only direct contributors to most of it, I'm fine with CC-By-SA, and I think he will be too (but I'll confirm); but a lot of the technical content was developed as I read the Bitcoin Core source code and could, I think, be considered a derived work.

I recommend doing what we did on Bitcoin.org, which is let each file's header reference its license (or have a COPYING file in directories of images or other file types where license metadata is hard to set).  If we have a set of approved licenses which are compatible with each other, such as CC-By-SA being compatible with MIT, and only allow content under those licenses, I don't think this should create problems.

CONTRIBUTING.md
btcdrak: "Fair use" rules apply in those circumstances.

CONTRIBUTING.md
```diff
@@ -27,3 +27,14 @@ and final decision is with the Bitcoin Core project lead.
...
+## License
+
+The code and documentation of this website are licensed under [MIT license][MIT] and all contributors agree to license their
+contributions under the same license.
+
+Unless specified in the header of the file, the content of this website is licensed under the Creative Commons Attribution-ShareAlike 4.0 International License: [CC-By-SA]
+and all contributors agree to license their content under the same license.
+
+[MIT]: https://github.com/bitcoin-core/website/blob/gh-pages/LICENSE
+[CC-By-SA]: http://creativecommons.org/licenses/by-sa/4.0/
```
MarcoFalke: Nit: CC-BY-SA

CONTRIBUTING.md
```diff
@@ -27,3 +27,14 @@ and final decision is with the Bitcoin Core project lead.
 
 Translations should be made from the English source. The English source has a `version:` field in the Front Matter which should be updated if on the rare
 occasion the English text is updated. Translated text should remove `translated: false` from the Front Matter.
+
+## License
+
+The code and documentation of this website are licensed under [MIT license][MIT] and all contributors agree to license their
```
MarcoFalke: Is it `documentation of this website` or `documentation on this website (i.e. Bitcoin Core documentation)`?

btcdrak:
Updated for @harding's request. This allows old content to be MIT for example. We don't have to accept PRs with varying demands.

PR #17, opened 2016-01-13
[blog] Add commit visualisation for 2015

btcdrak:
cacrqdp

### DIFF
--- a/_posts/en/posts/2016-01-13-developer-visualisation-2015.md
+++ b/_posts/en/posts/2016-01-13-developer-visualisation-2015.md
@@ -1,0 +1,28 @@
+---
+layout: post
+id: en-development-visualisation-2015
+name: development-visualisation-2015
+title: Core Development Visualisation for 2015
+permalink: /en/2016/01/13/development-visualisation-2015
+version: 1
+---
+The following video shows commit activity in the [Bitcoin Core repository][repository] during 2015. A full list of code contributors during this period can be found [here][activity].
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/3861hT0wY2U" frameborder="0" allowfullscreen> </iframe>
+
+In 2015, the Bitcoin Core project released 2 major versions of it's software together with a further 5 maintenance releases. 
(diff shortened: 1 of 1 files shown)


CodeShark:
ACK

jmcorgan:
Don't have the time atm to render locally, so utACK. Might want to explain just a bit more the benefits of the new features, but it is good as-is.

jonasschnelli:
Hmm.. not sure, tend to NACK. First, I don't see a big value in adding the visualization. Second: the music is just so cheesy.

MarcoFalke:
Oh, and the music may be copyrighted. I don't understand copyright well enough to understand what this means for us but I'd rather not risk getting a DMCA takedown.

btcdrak:
@MarcoFalke It's embedded content from Youtube.

MarcoFalke:
@jonasschnelli Would you ACK if the video was removed (or muted)?

jonasschnelli:
I think plain stats (amount of commits per month, per year), probably not user based (we don't want to raise a competition here) would be better. If website visitors could quick get the current activity rate (in commits-per month / comments per day / etc.) that would be nice IMO.

The visualization is just a gimmick IMO that has not much of a value.

adam3us:
ACK. maybe some stats later but publish ACK

btcdrak:
We can add more details after but I think we should go ahead and publish this now.

harding:
Can we leave more time for review next time?  I just finished watching, and the video has CLTV mixed up with Strict DER:

[Caption below says "July 2015" and "CLTV activation"]

screenshot 2016-01-13 at 09 19 03 - display 1

btcdrak:
@harding I contacted the author about this earlier this morning. He will update the annotations. @jonasschnelli is also going to make a version with avatars plus some extra charts.

PRabahy:
Oops, hadn't been on reddit in a few days. I'll fix the CLTV Activation annotation.

The music was something that was recommended and inserted by youtube itself, so I assume that youtube already figured out any copyright issues. If anyone is interested in doing more research it is Blue Collar by Gavin Froome.

Looks like the idea was redone at https://www.youtube.com/watch?v=FIt7GLxxIpY and I think I like the way those annotations were done a bit more.

PR #20, opened 2016-01-13
Fix excerpts

btcdrak:
Markdown link don't appear to be rendering with auto excerpts in the `/` or `/bog/`.

### DIFF
--- a/_posts/en/posts/2015-12-14-segregated-witness.md
+++ b/_posts/en/posts/2015-12-14-segregated-witness.md
@@ -8,6 +8,7 @@ author: sipa
 related: false
 tags: [scalability]
 version: 1
+excerpt: This is the extended presentation of Segregated Witness by Pieter Wuille.
 ---
 
 This is the extended presentation of Segregated Witness by Pieter Wuille.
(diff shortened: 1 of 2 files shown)


harding:
I believe the reason the auto except doesn't render is because the front page doesn't include the Markdown reference definition, i.e.:

    [repository]: https://github.com/bitcoin/bitcoin

We fix this class of problems on Bitcoin.org by having a global link definition file that gets sourced on most Markdown pages.  E.g.

    {% include references.md %}

Where `references.md` would include:

    [repository]: https://github.com/bitcoin/bitcoin

A faster hack for this specific problem would just use inline links in the except area:

    [Bitcoin Core repository](https://github.com/bitcoin/bitcoin)

I have tested this PR and can confirm that it also solves the problem.  One of the solutions mentioned above would be my preference, but this also gets my tested ACK.

btcdrak:
@harding I originally tried your last "faster hack" but it didnt work for me, I get the following on my local setup
ruegk3x

btcdrak:
We better merge this as an immediate fix since it affects the front page.

btcdrak:
@harding I also found problems with using `except_separator` so maybe something is broken on my local setup.

harding:
@btcdrak thinking about it some more, I think using the YAML definition is probably the best way.  Although making the first paragraph the excerpt is convenient, I think you usually don't want to spoil your lede by making it your teaser.

Also, I should've said it in my post above, but I was perfectly fine with fast merging this to fix the bug.  Thank you for doing that.

PR #24, opened 2016-01-13
Use full width (reduce right margin to 12/0.5)

jonasschnelli:
IMO using the full page with makes things more readable.

Before:
bildschirmfoto 2016-01-13 um 21 55 45

After:
bildschirmfoto 2016-01-13 um 21 56 01

### DIFF
--- a/_sass/page.scss
+++ b/_sass/page.scss
@@ -239,12 +239,12 @@ $button-size: 1.5rem;
 		@include suffix(12,1);
 		margin-bottom: 2em;
 		@media #{$small} {
-			@include grid(12,6);
+			@include grid(12,8.5);
 			@include prefix(12,0);
-			@include suffix(12,0);
+			@include suffix(12,0.5);
 		}
 		@media #{$x-large} {
-			@include grid(12,4.5);
+			@include grid(12,7);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


MarcoFalke:
Not sure if we want the text go all the way to the right and have the logo on the left dictate the margin...

jonasschnelli:
The idee is, that the page width should be flexible. I always disliked the fixed width of https://bitcoin.org/en/developer-guide#block-chain:

harding:
Although my understanding of typesetting is admittedly poor, it was my understanding that paragraph text should be fairly narrow to make it easy for the reader to go from the right end of one line to the left start of the next line without losing their place, as is easy to do when the lines are long.

You can see the pervasiveness of this typesetting rule here on GitHub, where the comments are about 600px wide, or on designed-for-reading sites like Medium where again the text is limited to about 600px, or on lots of other well-designed sites.

Now, in the Bitcoin.org developer documentation, we have lots of tables and illustrations that would certainly benefit from additional margin space, but one reason I've chosen to keep them cramped is because aiming for 600px means the images and tables work fine on most tablets and even when scaled down slightly to fit smartphones in profile mode.

That said, I'm ok with this change if other people want it.  I've tested it and didn't notice any technical problems on the English pages.

dexX7:
At least on a bigger screen I prefer narrow text.

MarcoFalke:
Also, it looks odd, if you don't use the space on the left. I know there is a jpeg hidden but this would be a good time to get rid of that.

screenshot from 2016-01-18 09-35-29

PR #31, opened 2016-01-15
Switch domain name

btcdrak:
This will be merged on January 15th after 15:00UTC

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,1 @@
-bitcoinco.re
+bitcoincore.org


luke-jr:
We can't use both?

btcdrak:
@luke-jr bitcoinco.re will be redirected to bitcoincore.org for the time being, certainly cant have duplicated content but still could use the bitcoinco.re domain for other things. People just prefer the .org

harding:
If I recall correctly, there was other stuff previously hosted on BitcoinCore.org, such as downloads for the Gitian compile and some fee analysis stuff by Gavin.  What has happened to that?

btcdrak:
@harding moved to a subdomain. The gitian stuff is only a backup URL.

harding:
@btcdrak cool.  Then, ACK.

PR #37, opened 2016-01-15
Initial Software EOL policy

btcdrak:
Todos:

- [x] add to menus
- [x] get exact dates from @laanwj 

ehm0h2j


harding:
I think the text about only backporting features to the point releases is not quite correct.  My understanding is that we're trying to *introduce* all new soft forks in point releases (i.e. not in major versions like we did for BIP66).  I suggest saying:

> We will provide maintenance "point releases" that provide important bug fixes or new features that cannot wait for the next major version or which we think are better deployed in a minor version where people can adopt them quickly without having to analyze a long changelog (for example, soft-forking code).  Point releases will be numbered 0.11.1, 0.11.2, 0.12.1, 0.12.2 etc.

luke-jr:
Maintenance end is not a set point. It's more of only soft-guaranteeing they will be maintained until that point, but we may choose to maintain them longer.

btcdrak:
@harding It might be better to to focus on what happens procedurally. Consensus changes are always developed in the master branch first. Once they are ready, they are backported to maintenance branches which we will then release as deployable software. The same holds true for bugs which we feel must be released into maintained versions. The fact that we like to introduce softforks in maintained versions first is orthogonal to the process. We do it because it allows primarily allows enterprise users who have different strict QA process for upgrades.

I like your text but I think we also need to be careful since technically maintenance version will receive mostly bug fixing and sometimes minor feature additions, as well as soft-forks (primarily because it increases the speed of adoption). We need to be careful not to imply that people should _expect_ new features in the maintenance versions.

@luke-jr Agree, that is why there's a gap between maint.end and EOL where we would guarantee critical security fixes but doesn't forbid us from adding other fixes at our discretion. The general rule of thumb so far is "maint.end" is shortly after the next major release. @laanwj of course needs to clarify the exact details.

MarcoFalke:
I think 0.8 is already EOL due to being vulnerable to the upnp thing?

luke-jr:
@btcdrak The phrasing sounds more absolute. But for example, 0.10 will continue to get full support after @laanwj moves on, since I will be taking over maintaining it for a [as of yet undetermined] time.

luke-jr:
FWIW, 0.8.x de facto ended support 2013-12-09 for v0.8.6.

For 0.9, 2015-05-20 (v0.9.5).

luke-jr:
@MarcoFalke Sounds likely; I think 0.8 and 0.9 were never patched to fix that.

MarcoFalke:
0.9 received the patch but I didn't check if a new tag was created for gitian builds.

laanwj:
New features are only in major versions (there have been some exceptions, but sometimes the line between bugfix and feature is blurred). Minor versions get bugfixes, translation updates, and softforks.

I tend to regard the last two major releases as maintained: so at this point 0.10 and 0.11, after the 0.12 release that will be 0.11 and 0.12. 

The older the major release, the more critical issues have to be to get backported to it, and the more to warrant a new minor release. E.g. in the case of upnp issue, it was backported to 0.9, but no one bothered doing a new minor release. Usually the assumption is that people on such releases will have their own patch sets (hence the upgrading inertia), so don't really benefit from binary releases.

But indeed, as @luke-jr says, it doesn't depend just on me. It's perfectly OK if luke-jr wants to adopt some release (such as 0.10.x) and maintain it for longer.

btcdrak:
OK I've updated preview image with the latest, taking into account @luke-jr and @laanwj 

/cc @harding

luke-jr:
What is TDB? Shouldn't it be TBD?

harding:
Luke, I believe TDB is the Yoda-speak version of TBD. :-)  (Or if you're really dorky, the Ciceroean version.)

Nitpick: I suggest that instead of "TBD", we write out "To be determined" or just "Unknown" so people don't have to ponder abbreviations they may not be familiar with.

btcdrak:
@harding Yoda! haha :smile:

I put a key at the bottom.

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -0,0 +1,28 @@
...
+
+We maintain the major versions until their "Maintenance End". We generally maintain the current and previous major release. So if the current release is 0.12, then 0.11 is also considered maintained. Once 0.13 is released, then 0.11 would be considered at it's "Maintenance End". The older the major release, the more critical issues have to be to get backported to it, and the more to warrant a new minor release. Once software has reached the "Maintenance End" period it will only receive critical security fixes until the EOL date. After EOL, users must upgrade to a later version to receive security updates. 
+
+Please note that minor versions get bugfixes, translation updates, and soft forks. Translation on Transifex is only open for the last two major releases.
+
+For example, major version 0.9 was released on 2014-03-19  and we provided maintenance fixes (point releases) until 2015-02-16. Critical security issues would still be continue to be fixed until the End-Of-Life "EOL" date of 2016-12-31. However, to take advantage of bug fixes, you would have to upgrade to a later major version.
```
MarcoFalke: Is this consistent with the table?

btcdrak:
Was discussed at meeting today http://bitcoinstats.com/irc/bitcoin-dev/logs/2016/01/21#l1453404632.0

MarcoFalke:
I missed the meeting but the tl;dr:

> 19:40:29 sipa> final release of 0.X means EOL of 0.(X-2)

_posts/en/pages/2016-01-15-lifecycle.md
MarcoFalke: @btcdrak ?

laanwj:
ACK

btcdrak:
I've talked this over with @laanwj and it's good enough. EOL is a living document and dates will have to be updates as we go along anyway.

PR #43, opened 2016-01-18
Fix related-articles (move from header to post area)

jonasschnelli:
**Before:**
bildschirmfoto 2016-01-18 um 08 52 33

**After:**
bildschirmfoto 2016-01-18 um 08 53 16


btcdrak:
tested ACK. Thanks, I'll merge this quickly since it's a major bug.

PR #45, opened 2016-01-18
Update 2016-01-13-supported-bips.md

MarcoFalke:
Add Bitcoin Core 0.12 because 0.11 does not support RBF.

### DIFF
--- a/_posts/en/pages/2016-01-13-supported-bips.md
+++ b/_posts/en/pages/2016-01-13-supported-bips.md
@@ -5,7 +5,7 @@ type: page
 permalink: /en/bips
 share: false
 ---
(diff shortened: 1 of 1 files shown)


btcdrak:
Maybe it would be better to add that after since the current official version in 0.11?

MarcoFalke:
An alternative would be to extend the table with another column and include the first version which supports it. Like https://github.com/bitcoin/bitcoin/blob/master/doc/bips.md

btcdrak:
@MarcoFalke That's a great idea. Would you like to make a PR?

MarcoFalke:
Sure, see #49

PR #52, opened 2016-01-18
update IRC meeting 2016-01-14

G1lius:


btcdrak:
Brilliant, thanks for this!

PR #64, opened 2016-01-23
Overhaul supporters page

jonasschnelli:
Testsite: https://coresitetest.jonasschnelli.ch/en/supporters/

Added some margin between the "Show your Support" element and the footer.
Broaden the supporters page a little bit, left / right align first and last col.


btcdrak:
As requested, thanks!

PR #69, opened 2016-01-25
Fix TOC top margin

### LINKED ISSUE #68: ToC overlaps with header navigation
The ToC of the RBF FAQ overlaps with the header navigation, and it looks a bit messy this way in my opinion:

overlap

jonasschnelli:
fixes #68


jonasschnelli:
Should look like...
bildschirmfoto 2016-01-25 um 15 32 57

dexX7:
utACK. looks better in my opinion.

PR #70, opened 2016-01-25
black/inverted mouse-over state for subnavigation

jonasschnelli:
Adds mouse-over state in subnavigation.

bildschirmfoto-2016-01-25-um-17 22 13

Testsite: https://coresitetest.jonasschnelli.ch


btcdrak:
Thanks!

PR #82, opened 2016-01-27
In-sentence line-break removal for Transifex compatibility 

mably:
To facilitate translating with Transifex, every sentence should be kept on one line only.

This PR removes two extra line-breaks present in two french version pages.

### DIFF
--- a/_posts/fr/pages/2016-01-01-about.md
+++ b/_posts/fr/pages/2016-01-01-about.md
@@ -12,7 +12,6 @@ Bitcoin Core is an [open source](https://opensource.org/) project which maintain
 
 It is a direct descendant of the original Bitcoin software client released by Satoshi Nakamoto after he published the famous [Bitcoin whitepaper](https://bitcoin.org/bitcoin.pdf).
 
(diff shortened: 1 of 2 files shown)


MarcoFalke:
Concept ACK

btcdrak:
This doesnt matter now, because we're taking the content from the English and this file will be overwritten when Transifex is translated.

MarcoFalke:
Sure, it would need adjustment in the English source, I guess?

btcdrak:
@MarcoFalke Yes. So long as `en` has no line breaks mid sentence then all is good for Transifex.

mably:
@btcdrak you asked me to make this PR this afternoon on slack, did I missed something?

btcdrak:
@mably We only need to worry about the English. The rest comes from Transifex.

mably:
Ok I missed that part I guess, do you want me to redo it for the corresponding english pages?

btcdrak:
I think I fixed it earlier. Please check.

mably:
@btcdrak nope, these two line-breaks are still present in the original english files.

btcdrak:
Then PR the English.

mably:
Ok

PR #83, opened 2016-01-27
In-sentence line-break removal for Transifex compatibility

mably:
To facilitate translating with Transifex, every sentence should be kept on one line only.

This PR removes one extra line-break present in the 2016-01-01-about.md english page.


btcdrak:
This should auto update Transifex.

PR #93, opened 2016-01-31
Meetings page translations fix

mably:
Translations were not correctly displayed in meetings page.


btcdrak:
I dont understand the difference.

mably:
Looks like the loop variable *post* can't be overriden.

In fact previous version doesn't show translated meeting pages.

I took example on the home page that works for translated posts.

btcdrak:
ACK

PR #107, opened 2016-02-04
[EOL] Update _maintenance-table.md

MarcoFalke:
There was no 6 month release cycle for 0.9 and earlier.

* Support for 0.8 was dropped in Nov 2013 (with 0.8.6).
* Support for 0.9 was dropped in Jul 2015 after the release of 0.11.1 and the last minor version in May 2015 (0.9.5)


btcdrak:
I disagree. Maintenance support and EOL are not the same. Maintenance period is Current release 0.X where maintained versions are 0.X and 0.X-1. However, that _does not_ end support entirely: _critical_ security issues will continue to be back-ported even if no binaries are produced until the EOL date. EOL is the date at which we are declaring, "if there are any problems with that version beyond that date, you must upgrade to a newer version to get the fix, even if it's a critical security issue". Since critical security issues are extremely rare, this is not a burden of development, and has been the status quo

0.8.6 was tagged on 2013-12-09, maintenance end was 2014-03-19 and received it's last security patch on 2015-02-17. 

0.9.5 was tagged on 2015-05-21, with last commit 2015-10-15. 

I also disagree with having rough months, over specific dates.

MarcoFalke:
There seems to be some misunderstanding. I will try to reach you via IRC in the coming weeks.

btcdrak:
@MarcoFalke OK.

MarcoFalke:
a779f2381dda088cbd23033f0ab62bec45960495 looks good. Thanks @btcdrak

PR #129, opened 2016-03-01
Correct 2016-02-25 meeting links

veqtrus:
I swapped the logs and meetbot links. Also added the BIP 109 link.

Sorry for not noticing earlier.


MarcoFalke:
ACK

PR #138, opened 2016-03-15
Meeting summaries: 2016-03-10

harding:
Last week's meeting summary. A preview may be seen here: http://dg0.dtrt.org/en/meetings/2016/03/10/ 

I believe @G1lius should be back for this Thursday's (3/17) meeting.

@btcdrak regarding your comment from last week about separate subheads for background, meeting comments, and conclusion to make the notes more skimable, I think it's not good typesetting to have subheads with fewer than three paragraphs between them so I didn't add them.  However, this week I did use standout formatting to try to accomplish the same thing: background is placed in italics; meeting comments are unstyled; and the conclusion begins with the phrase "Action item(s):" in bold.

As always, please let me know if there's anything I can improve.  Thanks!

### DIFF
--- a/_includes/_references.md
+++ b/_includes/_references.md
@@ -8,6 +8,11 @@
 [Cory Fields]: https://github.com/theuni
 [Alex Morcos]: https://github.com/morcos
 [Pavel Janik]: https://github.com/paveljanik
+[Matt Corallo]: https://github.com/TheBlueMatt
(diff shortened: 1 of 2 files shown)


G1lius:
Thanks @harding   
Is everyone happy with this formatting? If so I'll continue in this trend.

Still need to catch up with everything though.

btcdrak:
@G1lius I prefer the old format.

PR #140, opened 2016-03-18
Fix typo in segwit dev guide

achow101:
Fixed a typo in the segwit wallet dev guide where it said that p2wsh outputs used ripemd160(sha256(witnessScript)) instead of the actual sha256(witnessScript)


btcdrak:
Good catch. Confirmed with @CodeShark.

PR #142, opened 2016-03-21
Summary for 2016-03-17

G1lius:
I've done the old style of background-comments-conclusion.
Comments welcome, been out of the loop for a month so I might have missed information.


btcdrak:
Render checked.

I'll fix it, but can you make sure at full sentences are are on one line. This is because of how Transifex works, breaking up sentences on multiple lines makes it untranslatable.

laanwj:
ACK, thanks

PR #162, opened 2016-04-21
Update 2016-01-18-segwit-wallet-dev.md

voisine:
BIP141 states: "The only item in scriptSig is hashed with HASH160" under the section titled: "Witness program nested in BIP16 P2SH"


jl2012:
ACK

PR #179, opened 2016-06-08
Update lifecycle page

btcdrak:
There has been repeated confusion about Bitcoin Core's release versioning compared with SemVer 2.0.0.

While Bitcoin Core does indeed follow the spirit of SemVer regarding major and minor releases, where minor releases focus on bug fix maintenance; Bitcoin Core software releases are decoupled network versioning and consensus rule changes. As such, the exceptions need to be explained explicitly.

### DIFF
--- a/_posts/en/pages/2016-01-15-lifecycle.md
+++ b/_posts/en/pages/2016-01-15-lifecycle.md
@@ -58,14 +58,12 @@ _TBA: to be announced_
 The description above only describes Bitcoin Core software releases. Many other parts of the Bitcoin system contain their own versions.  A few examples:
 
 - Every **transaction** contains a version number.
-- Every **block** allows miners to signal their readiness to enforce new
-  versions of the consensus rules.
 - The **P2P network protocol** uses version numbers to allow nodes to announce what features they support.
 - Bitcoin Core's **built-in wallet** has its own internal version number.
 
 These versions numbers are deliberately decoupled from Bitcoin Core's version number as the Bitcoin Core project either has no direct control over them (as is the case with blocks and transactions), or tries to maintain compatibility with other projects (as is the case with the network protocol), or allows for the possibility that no major changes will be made in some releases (as is sometimes the case with the built-in wallet).
 
-The consensus protocol itself doesn't have a version number, although it has gone through about a dozen different versions since the release of Bitcoin 0.1 in January 2009.
+The consensus protocol itself doesn't have a version number.
 
 ## Relationship to SemVer
 
--- a/_posts/en/pages/2016-01-15-lifecycle.md
+++ b/_posts/en/pages/2016-01-15-lifecycle.md
@@ -55,20 +55,13 @@ _TBA: to be announced_
 
 ## Protocol versioning
 
-The description above only describes Bitcoin Core releases.  Many other
-parts of the Bitcoin system contain their own versions.  A few examples:
+The description above only describes Bitcoin Core software releases. Many other parts of the Bitcoin system contain their own versions.  A few examples:
 
 - Every **block** and **transaction** contains a version number.
-- The **P2P network protocol** uses version numbers to allow nodes to
-  announce what features they support.
+- The **P2P network protocol** uses version numbers to allow nodes to announce what features they support.
 - Bitcoin Core's **built-in wallet** has its own internal version number.
 
-These versions numbers are deliberately decoupled from Bitcoin Core's version
-number as the Bitcoin Core project either has no direct control over them (as is
-the case with blocks and transactions), or tries to maintain compatibility with
-other projects (as is the case with the network protocol), or
-allows for the possibility that no major changes will be made in some
-releases (as is sometimes the case with the built-in wallet).
+These versions numbers are deliberately decoupled from Bitcoin Core's version number as the Bitcoin Core project either has no direct control over them (as is the case with blocks and transactions), or tries to maintain compatibility with other projects (as is the case with the network protocol), or allows for the possibility that no major changes will be made in some releases (as is sometimes the case with the built-in wallet).
 
 ## Relationship to SemVer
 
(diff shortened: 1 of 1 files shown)


_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -7,20 +7,42 @@ layout: page
...
+These will be numbered 0.11.0, 0.12.0 etc.
+
+## Maintenance releases
 
-We will provide maintenance "minor releases" that fix bugs within the major releases. We may add minor features to point releases where necessary, for example when back-porting consensus rule changes such as soft forks. Minor releases will be numbered 0.11.1, 0.11.2, 0.12.1, 0.12.2 etc.
+We will provide maintenance "minor releases" that fix bugs within the major releases. As a general rule we do not introduce major new features in a maintenance release (consensus rule are exempt from this rule). However, we may add minor features where necessary, and we will back-port consensus rule changes such as soft forks.
+
+Minor releases will be numbered 0.11.1, 0.11.2, 0.12.1, 0.12.2 etc.
+
+## Soft forks
```
luke-jr: No reason to think the same wouldn't apply to hard forks as well (actually, it's even more important for HFs than SFs..). Maybe we should just refer to "consensus rule changes" here?

luke-jr:
Looks like a good opportunity to update that 0.11 is no longer supported, and we can't really support more than the most recent.

MarcoFalke:
Concept ACK fc88445

btcdrak:
ping @harding

harding:
Opened a [PR to @btcdrak's branch](https://github.com/btcdrak/bitcoincore.org/pull/2) with a few nits and the following revised final section:

> ## Relationship to SemVer

> Bitcoin Core software versioning is similar to the [SemVer][] optional versioning standard for regular feature releases.  However, SemVer was designed for use in normal software libraries where individuals can choose to upgrade the library at their own pace, or even stay behind if they don't like the changes.

> Parts of Bitcoin, most notably the consensus rules, don't work that way.  In order for a new consensus rule to go into effect, it must be enforced by some number of miners, full nodes, or both; and once it has gone into effect, software that doesn't know about the new rule may generate or accept invalid transactions.

> For this reason, Bitcoin Core deviates from SemVer for changes to consensus rules and other updates where network-wide adoption is necessary or desirable.  The deviation is slight: Bitcoin Core releases these changes as minor releases (`x.y`) instead of as major releases(`x.0`); this minimizes the size of the patch in order to make it easy for as many people as possible to inspect it, test it, and deploy it.  It also makes it possible to backport the same patch to multiple previous major releases, further increasing the number of users who can easily upgrade, although there are not always enough volunteers to manage this.

[SemVer]: https://semver.org/

As always, please feel free to reject any changes you don't like.

laanwj:
ACK on @btcdrak's text and @harding's improvements

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -25,13 +25,13 @@ These will be numbered 0.11.0, 0.12.0 etc.
...
 ## Maintenance releases
 
-We will provide maintenance "minor releases" that fix bugs within the major releases. As a general rule we do not introduce major new features in a maintenance release (consensus rules are exempt). However, we may add minor features where necessary, and we will back-port consensus rule changes such as soft forks.
+We will provide maintenance "minor releases" that fix bugs within the major releases. As a general rule we do not introduce major new features in a maintenance release (except for consensus rules). However, we may add minor features where necessary, and we will back-port consensus rule changes such as soft forks.
 
 Minor releases will be numbered 0.11.1, 0.11.2, 0.12.1, 0.12.2 etc.
 
 ## Consensus rules
 
-Proposals to change consensus rules are always shipped first in maintenance versions such as 0.11.2, 0.12.1 etc. This make it easier for enterprise users to assess and test the proposal due to smaller changeset than in a major release. It also allows users who follow a more conservative upgrade path to adopt consensus rule changes in a more timely manner.
```
CyrusV5: 'This make it' -> 'This makes it'.

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -53,9 +53,23 @@ Once EOL is reached, you will need to upgrade to a newer version.
 
 _TBA: to be announced_
 
-## Foot notes
+## Protocol versioning
 
-Bitcoin Core software versioning is similar, but not identical to [SemVer][]. Bitcoin Core implements a full stack of wallet, full node, networking and consensus rule validation. However, versioning of the Bitcoin system is much more complex; there are block versions, transaction versions, p2p versions, and consensus rules are all versioned separately, and the consensus rule. [SemVer][] is not designed to handle this complexity and while Bitcoin Core adheres in the most part to the spirit of [SemVer][], there are notable exceptions, especially for consensus rules proposals.
+The description above only describes Bitcoin Core software releases. Many other parts of the Bitcoin system contain their own versions.  A few examples:
+
+- Every **block** and **transaction** contains a version number.
```
luke-jr: Blocks don't really since BIP 9... Even if we interpret the bits as a number (a stretch in itself), they indicate future changes, nothing about the block itself.

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -53,9 +53,23 @@ Once EOL is reached, you will need to upgrade to a newer version.
...
+- The **P2P network protocol** uses version numbers to allow nodes to announce what features they support.
```
luke-jr: "might support"?

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -53,9 +53,23 @@ Once EOL is reached, you will need to upgrade to a newer version.
...
+- Bitcoin Core's **built-in wallet** has its own internal version number.
+
+These versions numbers are deliberately decoupled from Bitcoin Core's version number as the Bitcoin Core project either has no direct control over them (as is the case with blocks and transactions), or tries to maintain compatibility with other projects (as is the case with the network protocol), or allows for the possibility that no major changes will be made in some releases (as is sometimes the case with the built-in wallet).
```
luke-jr: The wallet versioning is more of a case where we don't break backward compatibility unless the user explicitly opts in.

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -53,9 +53,23 @@ Once EOL is reached, you will need to upgrade to a newer version.
 
 _TBA: to be announced_
 
-## Foot notes
+## Protocol versioning
```
luke-jr: This talks about the literal numbers, but doesn't mention the actual consensus protocol itself, which (for better or worse) is no longer indicated by any given number.

_posts/en/pages/2016-01-15-lifecycle.md
harding: I think the "allow" in that sentence communicates the same thing as "might".

_posts/en/pages/2016-01-15-lifecycle.md
harding: How about this?:

- Every **Transaction** contains a version number.
- Every **block** allows miners to signal their readiness to enforce new versions of the consensus rules.

_posts/en/pages/2016-01-15-lifecycle.md
harding: Is that actually used?  I can't think of any release where I had to explicitly opt-in to a wallet change; I just made a backup before switching to the new version, started the new version, and it worked.

Moreover, I don't know of any way to check the wallet version without starting bitcoind/bitcoin-qt (since you have to run an RPC to get the wallet version number).

btcdrak: Overall the whole point to get across is that Bitcoin Core and the rest of the system are decoupled and SemVer simply not cannot cope with this.

_posts/en/pages/2016-01-15-lifecycle.md
btcdrak: fixed

_posts/en/pages/2016-01-15-lifecycle.md
btcdrak: Well there isnt consensus versioning...

harding: Ok.  I'll add another short paragraph to the bottom of this section stating that.

_posts/en/pages/2016-01-15-lifecycle.md
btcdrak: This is pedantry. We know nodes can lie about the services they support but that's out of scope of this document.

harding:
@btcdrak and I made some minor edits based on @luke-jr's feedback.  I think the version as of daa79da is probably good enough to put out there; we can revise it or expand it again in a few months if there is any further confusion that we notice on the forums or social media.

btcdrak:
@harding OK let's publish. Thanks for your help!

PR #199, opened 2016-08-05
Release 0.13.0: Add release notes and announcement blog post

harding:
Similar to previous releases, this PR adds the following two new pages to the site:

- A new page with release notes from Bitcoin Core source tree (verbatim except for a Markdown formatting change described below)  
- A new blog post announcing the release and describing some of its many improvements.  Feedback appreciated.   [Preview](http://imgur.com/a/72zIx)

### Todo

**Edit 2016-08-22:** If 0.13.0 is released on August 23rd, then both of the todos below are done.

- [x] Update dates in this PR to the actual release date
    - `_posts/en/posts/2016-08-23-release-0-13-0.md` should have its filename and the `permalink` field in its YAML header changed to the actual release date
    - `_posts/en/releases/2016-08-23-release-0.13.0.md` should have its filename changed to the actual release date
- [x] If necessary, update `_posts/en/releases/2016-08-23-release-0.13.0.md` to the current release notes.  It is currently on https://github.com/bitcoin/bitcoin/commit/a402396dce64c42ea73535b7dde4a9164d430438
    - If you update, you can just paste the raw contents of the release notes from the Bitcoin Core source tree into the file below the YAML header.  After doing that you need to make the following substitution over the whole file to fix a difference between GitHub-flavor Markdown and whatever we use on BitcoinCore.org: `s/- #/- \\#/`

I'll try to keep on top of both of the above todos so that this stays ready for merge when 0.13.0 is released.


_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+permalink: /en/2016/08/05/release-0.13.0/
+type: posts
+layout: post
+share: true
+version: 1
+---
+{% include _toc.html %}
+{% include _references.md %}
+
+We're pleased to announce the official release of Bitcoin Core 0.13.0.  During the six-month development cycle leading to this release, [dozens of contributors][release notes contributors] have made [hundreds of notable improvements][release notes changelog] to Bitcoin Core.  Among the many upgrades available in this release, the following may be especially interesting to miners, node operators, and wallet users:
```
sipa: The links don't seem to work in the github rendered view (https://github.com/bitcoin-core/bitcoincore.org/blob/4d1897a01e4f136d2916d248c3c71d0d78479999/_posts/en/posts/2016-08-05-release-0-13-0.md), but I assume that is intentional.

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+
+- **Preparation for segregated witness** to increase capacity, eliminate unwanted transaction malleability, and enable new ways to upgrade Bitcoin's Script language using soft forks.  The code in this release prepares for segwit only; it does not activate segwit on mainnet and will not behave differently when segwit is activated.
```
sipa: "will not behave differently when segwit is activated" sounds confusing. What about "It does not support segwit on mainnet, so upgrading to a future version will be needed"?

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+
+**More information:**
+
+- [Release notes][release notes segwit]
+
+- [Segregated Witness: the next
+  steps](https://bitcoincore.org/en/2016/06/24/segwit-next-steps/)
+  BitcoinCore.org blog post
+
+- [BIP141: Segregated witness (consensus
```
sipa: Links to BIP 143, BIP 144, BIP 145?

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Compact block relay
+
+Prior to Bitcoin Core 0.13.0, a running full node would (by default) receive many transactions twice:
+
+1. Before the transaction was confirmed, as an individual transaction being relayed across the network.
+
+2. After the transaction was confirmed, as part of a group of transactions contained in a newly-mined block being relayed across the network.
+
+There's no need for the node to receive the transaction a second time if it still has the first copy, so compact block relay (BIP152) allows a node to receive from its peers an ordered list of what transactions are included in a new block so that the node can use the transactions it has already received to partly or fully reconstruct the transactions part of the block for itself.  If the node doesn't receive all the transactions it needs to fully reconstruct the block, it requests the missing transactions from its peers, and then uses them to complete the block.
```
sipa: That's a pretty long sentence.

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+
+![Compact Blocks diagram](https://raw.githubusercontent.com/bitcoin/bips/master/bip-0152/protocol-flow.png)
+
+Compact blocks provides two very important benefits to the network:
```
sipa: "3. Faster propagation of blocks across the network, by dramatically reducing the amount of data that needs to be transported at the time a block is found." ?

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+
+2. By eliminating the bandwidth spike that occurs when nodes receive a new block, compact blocks may make it easier to keep a node operating on connections with limited peak bandwidth.  For example, several users have reported that receiving a new block slows down other important activity on their network, such as video conferencing, so some of those users shut down Bitcoin Core before starting those activities.  Compact blocks may eliminate those bandwidth spikes and make running Bitcoin Core less inconvenient for those users.
+
+**More information:**
+
+- [Release notes][release notes compact blocks]
+
+- [Compact Block Relay FAQ](/en/2016/06/07/compact-blocks-faq/)
+
+- [BIP152][], which describes the compact blocks protocol
```
sipa: This link does not work?

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+
+Prior to Bitcoin Core 0.13.0, nodes didn't tell each other what minimum fee rate they were using, which could lead to wasted bandwidth.  For example, Alice sends Bob a transaction without realizing that the transaction's fee rate is below Bob's minimum.  Because of the way Bitcoin transactions are relayed, Bob has no way of knowing that the transaction is below his limit until he has already downloaded the whole transaction, at which point he stops processing the transaction because its fee rate is too low, so both his bandwidth and Alice's bandwidth end up being wasted.
+
+Bitcoin Core 0.13.0 supports a new message that has been added to the peer-to-peer (P2P) protocol, the `feefilter` message, which has been designed to help eliminate this wasted bandwidth. This P2P message allows Bob to tell Alice what he is currently using as his minimum relay fee rate so that Alice will not bother trying to relay to him any transactions that pay a fee below his rate.
```
sipa: Link to BIP130?

sipa:
Nice work! Only a few nits above.

_posts/en/posts/2016-08-05-release-0-13-0.md
btcdrak: It should pick up from the _references.md

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,209 @@
...
+version: 1
+---
+{% include _toc.html %}
+{% include _references.md %}
+
+We're pleased to announce the official release of Bitcoin Core 0.13.0.  During the six-month development cycle leading to this release, [dozens of contributors][release notes contributors] have made [hundreds of notable improvements][release notes changelog] to Bitcoin Core.  Among the many upgrades available in this release, the following may be especially interesting to miners, node operators, and wallet users:
+
+- **Preparation for segregated witness** to increase capacity, eliminate unwanted transaction malleability, and enable new ways to upgrade Bitcoin's Script language using soft forks.  The code in this release prepares for segwit only; it does not activate segwit on mainnet and will not behave differently when segwit is activated.
+
+- **Compact block relay** on the peer-to-peer network to eliminate a major source of redundant data transfer among nodes that relay transactions, as well as reduce the peak amount of bandwidth those nodes use when downloading newly-generated blocks.
```
instagibbs: perhaps mention propagation speedup as well

_posts/en/posts/2016-08-05-release-0-13-0.md
instagibbs: + "and opportunistically sending small amounts of data to peers that ask for it"?

instagibbs:
LGTM aside from all the links.

harding:
Pushed commit https://github.com/bitcoin-core/bitcoincore.org/pull/199/commits/751492941513758f37b56cb128a1b01ad516967a which I believe resolves all feedback from @sipa and @instagibbs .  Also pushed commit https://github.com/bitcoin-core/bitcoincore.org/pull/199/commits/10bb8e2b42b844131f232a09f0aff269ba7219d2 which updates the 0.13.0 release notes to the same content as https://github.com/bitcoin/bitcoin/commit/a402396dce64c42ea73535b7dde4a9164d430438 .

I also rendered the page locally and verified that all reference-style links (`[text][reference]`) were correctly rendered into hyperlinks.

Thanks for the reviews!

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,210 @@
...
+
+- [Compact Block Relay FAQ](/en/2016/06/07/compact-blocks-faq/)
+
+- [BIP152][], which describes the compact blocks protocol
+
+- [Bitcoin FIBRE](http://bitcoinfibre.org/), an open source protocol and implementation that builds upon compact blocks to minimize latency for new block announcements between peers on managed networks.  FIBRE was designed and implemented alongside compact block relay version 1 and it is being used to test improvements for a subsequent version of compact block relay.
+
+## Fee filtering
+
+For several years now, Bitcoin Core nodes have used a minimum relay fee rate to determine what unconfirmed transactions they'll process, relay, and store in their individual memory pools.  Each node gets to decide its own minimum relay fee rate, and if they receive a transaction whose fee rate is below that limit, they don't add it to their memory pools or relay it to their other peers.
```
sdaftuar: I think the language in this paragraph and the next are slightly confusing.  The min relay fee that Bitcoin Core nodes have historically had isn't a true minimum, in that free transactions would still get relayed (though rate-limited) by default.

It's only due to the mempool limiting introduced in 0.12 that causes nodes to get into a state where no new transactions are accepted below some "minimum mempool fee" (which is a function of the minimum relay fee, the transactions which have been evicted from the mempool, and how much time has passed).

Specifically, it's the minimum mempool fee that is broadcast in fee filter messages.  If the mempool fee is not in effect, eg because the mempool isn't full and hasn't been for a while, then no feefilter messages would be sent, regardless of what the `-minrelaytxfee` is.

jonasschnelli:
ACK the HD wallet part. The only thing that's missing is the possible xpriv export over dumpwallet, but not urgently required.

_posts/en/posts/2016-08-05-release-0-13-0.md
```diff
@@ -0,0 +1,210 @@
...
+
+- [Release notes][release notes hd wallet]
+
+- [Deterministic
+  wallets](https://en.bitcoin.it/wiki/Deterministic_wallet) (Bitcoin
+  Wiki)
+
+- [BIP32: Hierarchical Deterministic Wallets][BIP32]
+
+## Child pays for parent mining
```
sdaftuar: I don't think we should think of this as "child pays for parent mining", but instead as "ancestor fee rate mining", or something akin to that (which highlights the actual transaction selection algorithm).

In my view, "child pays for parent" is a wallet behavior, not a mining algorithm.  The mining algorithm introduced in 0.13 generally optimizes miner revenue better than the old algorithm, regardless of user behavior.

It is of course the case that wallets engaging in behaviors like "child pays for parent" -- ie putting a higher fee on a transaction that has an unconfirmed, lower fee parent -- should benefit from miners using 0.13.0  being smart enough to mine such transactions when optimal to do so.  But the right way of thinking about the mining algorithm is not in terms of particular user behavior we want to pick out; we should generally be thinking about how to make the mining code more optimal in general, and robust to any kind of user behavior.

So my suggestion for this section would be to describe the mining algorithm more along the lines of how it actually works (which could be as simple as saying it considers transactions for inclusion after sorting them by their feerate with unconfirmed parent transactions).  And then only use the term CPFP to describe wallet behaviors that will now work better as a result of this change.

In particular I'd strike the word "mining" from the title of this section (so it's just "Child pays for parent").

_posts/en/posts/2016-08-05-release-0-13-0.md
harding: @sdaftuar

> The min relay fee that Bitcoin Core nodes have historically had isn't a true minimum, in that free transactions would still get relayed (though rate-limited) by default.

I've added commit 2d0d7aff07535359155fb822752377abf8d19027 to mention tx priority, which hopefully clears up this inaccuracy.

> It's only due to the mempool limiting introduced in 0.12 that causes nodes to get into a state where no new transactions are accepted below some "minimum mempool fee"

If I understand your comment correctly, I think it's only correct if we assume that every node uses the same minimum relay fee.  Prior to 0.12, I almost always customized my minrelayfee, and I believe in one of the 0.11.x releases it was recommend for low-memory users to customize their minrelayfee (and, IIRC, in another 0.11.x, the default minrelayfee was temporarily increased).  This leads me to believe that even if I'm the only weirdo who customized his minrelayfee policy that historically a significant number of nodes were connecting to peers who had different minrelayfees and so bandwidth was being wasted when a node with a higher minrelayfee setting relayed transactions to a peer with a lower minrelayfee setting.

I wrote this section with that in mind, generalizing the situation across all modern releases to help keep it both correct, simple, and short for this announcement blog post which is almost certain too long already. :-)  I understand that the fixed-size mempools introduced in 0.12.0 with their variable minrelayfees aggravated the historic problem with wasted bandwidth and helped motivate the development of `feefilter`, but I'm not sure we need to go into that level of detail (but we can, if you think it's important).

If I've misunderstood your comment (or the technology), please let me know and I can make additional edits.  Thanks!

_posts/en/posts/2016-08-05-release-0-13-0.md
harding: @sdaftuar I push commit ba8717a4bcb97b3581a5708c491ae9774e0da3a2 which I think fully addresses your concerns with a minimum of edits.  Thanks!

harding:
I think this should be ready to go.  Filenames and metadata have been updated to assume a release date of 2016-08-23 UTC, and I've tried to address the most recent feedback.

_posts/en/posts/2016-08-05-release-0-13-0.md
sdaftuar: @harding Yeah I can appreciate that the nuance here is hard to get right for a general audience!  On further thought, I think your text is fine as-is -- the details here are probably not so important to anyone.

Since I started this comment though I might as well clarify the technology to make sure we're all on the same page: When an 0.13.0 node is started up, no fee filter messages are sent, no matter what you set your minrelaytxfee to.  It's not until the mempool fills up and some stuff gets evicted from it that a node will start sending feefilter messages.

So no bandwidth is saved until the mempool fills up.  Once the mempool does fill up, then it is true that the feefilter message will be generated, and its value will be at least as large as your minrelaytxfee (or higher, if transactions with feerates higher than your minrelayfee are getting evicted from the mempool).  But I'd imagine that if you set your minrelaytxfee to be very high, then you'd not save any bandwidth at all, because your mempool would never get full, so you'd never send a feefilter message.

Given that (a) mempools are generally full, and (b) we're even contemplating changing the behavior to be more along the lines of what you described anyway (maybe even for 0.13.1), the micro-detail about this only kicking in currently when mempools are full is not so important.

Thanks for writing this up!

sdaftuar:
@harding Looks good, thanks!

sipa:
Looks great, thanks!

btcdrak:
ACK from me.

Once the announcement is made I would also like to add the signed PGP release announcement both to the website, and linked to the ML. I can do that post merge.

kanzure:
https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2016-August/013055.html

PR #211, opened 2016-09-09
[0.11] Set EOL

MarcoFalke:
[Needs discussion]


btcdrak:
Where does this date come from?

MarcoFalke:
This is open for discussion, but I feel like we should define one in advance.

btcdrak:
@MarcoFalke let's bring this up at the next IRC meeting.

MarcoFalke:
Discussion:

The end-of-life “EOL” date for version 0.11 on the lifecycle page should be set. The maintenance date doesn’t reflect reality as backports to previous versions see no usage and get no review, thus developers have focused on current versions. Sipa tosses the idea of automatically showing a warning message X months after being released to warn users the version is no longer maintained.

https://bitcoincore.org/en/meetings/2016/09/15/

laanwj:
ACK

PR #221, opened 2016-09-28
Update 2016-01-15-lifecycle.md

MarcoFalke:
Mention community support on best effort basis.


btcdrak:
I personally disagree with removing EOL - we only just added EOL. With all fairness, most of the meeting discussion was people having discussion that were already made many months ago because they hadnt actually read this document which we agreed on months ago.

PR #231, opened 2016-10-14
Colourise segwit adoption table

ajtowns:
Here's a patch to add background colours to the segwit adoption table -- makes it a bit easier to spot which ones aren't done yet. Not sure if there's a meaningful difference between "wip" and "no"; I left "wip" as normal background but maybe it should be coloured "bad" too.

I couldn't get kramdown markup to add an attribute to cells, so switched to using liquid to generate HTML directly based on a CSV file, which seems fine. I chose yellow and blue because it seems to work for colour blind (unlike green and red), doesn't suffer from red is bad in the West and good in the East confusion, and because "golden" sounds positive and "blue" is sad, but whatever.

Also added a URL for NBitcoin since it was the only one missing a URL and it saved special casing it.

### DIFF
--- a/_data/segwitsupport.csv
+++ b/_data/segwitsupport.csv
@@ -1,0 +1,48 @@
+name,url,planned,ready,notes,depends
+AirBitz,https://airbitz.co/,yes,no,wallet,
+Armory,https://github.com/goatpig/BitcoinArmory/tree/SegWit/,yes,wip,wallet,bitcoin core
+Bcoin,https://github.com/bcoin-org/bcoin,yes,yes,library,
+Bitcoin India,https://bitcoin-india.org/,yes,wip,exchange,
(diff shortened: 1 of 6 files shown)


ajtowns:
Here's what it looks like for me fwiw:

image

btcdrak:
In most cases "wip" means it's been done, but is in a development branch or hasn't been merged.

btcdrak:
Maybe yes should be green.

ajtowns:
> Maybe yes should be green.

Green/blue is confusing to some people I think, http://www.color-blindness.com/2007/05/18/mixing-up-blue-and-green/ and at least 0x9fd (green) vs 0x9df (blue) performs pretty poorly on http://www.color-blindness.com/coblis-color-blindness-simulator/

Green/yellow (cfe/ffb) seems okay though: 
segwit-gy

btcdrak:
Yeah, that works too.

PR #67, opened 2016-01-24
post on segwit features

ajtowns:
Here's a post on the features of segregated witness *other* than the blocksize. Does it seem interesting for the new website?

It's a bit one-sided, in that it only lists benefits. I think maybe it would be better to include costs (and a corresponding "Who loses?") too? Thoughts?

PR #242, opened 2016-10-27
Add page with a technical overview of segwit costs and risk

ajtowns:
I said in the #67 PR "It's a bit one-sided, in that it only lists benefits. I think maybe it would be better to include costs too"; but at the time I couldn't work out how to say anything intelligent on the topic. Now that segwit is able to be deployed and activated, it seems appropriate to revisit that.

I've separated each section as a separate commit to maybe make things a little easier to review/comment on.

### DIFF
--- a/_posts/en/posts/2016-10-28-segwit-costs.md
+++ b/_posts/en/posts/2016-10-28-segwit-costs.md
@@ -115,3 +115,15 @@ full serialisation can be recovered from a simple flag indicating
 which format is in use (P2PKH, P2WPKH, P2WPKH/P2SH, etc) along
 with the actual data (the pubkey and signature).  (For example,
 [compacted_txn.txt](https://people.xiph.org/~greg/compacted_txn.txt))
+
+# Block validation costs
+
+With segwit, additional processing is introduced when validating
+a block in order both to check the witness merkle tree, and to
+deal with P2SH-encoded witness transactions. This requires about
+five additional SHA256 hashes per transaction, and an additional
+SHA256 per P2SH-encoded-P2WSH input, and an additional HASH160 per
+P2SH-encoded-P2WPKH output. This however only amounts to about six SHA256
+runs over at most 4MB of data, or roughly about 24MB of SHA256 data in
+total, which should translate to at most an additional 15s per block on
+a Raspberry Pi, or under a tenth of a second on more capable hardware.
--- a/_posts/en/posts/2016-10-28-segwit-costs.md
+++ b/_posts/en/posts/2016-10-28-segwit-costs.md
@@ -37,3 +37,81 @@ This post does not attempt to produce a conclusion as to whether the
 benefits outweigh the costs or whether segwit should be deployed or
 activated, but rather to assist by providing background information to
 assist stakeholders in making informed decisions.
+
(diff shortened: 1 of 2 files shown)


harding:
I read through this and thought it was very informative.  (Thanks for writing it, @ajtowns!)  I'm currently making some minor edits and will send them to AJ in a few hours as a PR to his branch.

_posts/en/posts/2016-10-28-segwit-costs.md
```diff
@@ -0,0 +1,621 @@
...
+    should remain equally effective, provided the monitoring tools are
+    able to track segwit spends at all.
+
+ 3. Ensuring miners mine valid blocks is obviously a high priority to
+    everyone involved, and significant work has
+    gone into guaranteeing this is the case with
+    segwit. This includes both the direct work, under
+    [BIP-145](https://github.com/bitcoin/bips/blob/master/bip-0145.mediawiki),
+    as well as indirect work, such as Compact Blocks
+    ([BIP-152](https://github.com/bitcoin/bips/blob/master/bip-0145.mediawiki)).
```
sipa: Link is to BIP145 instead.

harding: Note, this and a two other mislink errors is being corrected on my branch by using `_references.md` with reference style links for all BIP links that don't link to a specific document section.  (Just mentioning this here in case anyone sees them and so @ajtowns doesn't need to bother fixing them for the moment.)

_posts/en/posts/2016-10-28-segwit-costs.md
```diff
@@ -0,0 +1,621 @@
...
+drastically alter the consensus limits on blocks.
+
+## Simpler segwit
+
+Many of the benefits of segwit could logically be separated into
+independent changes, and evaluated and deployed separately. The
+implementation requirements for the various features are, however,
+closely related:
+
+ * Linear scaling of sighash operations: the CHECKSIG and CHECKMULTISIG
```
ajtowns: True. Maybe:

> Linear scaling of sighash operations: the CHECKSIG and CHECKMULTISIG opcodes need replacement, and some mechanism to maintain limits on the use of old CHECKSIG/CHECKMULTISIG.needs to be in place.

and changing the following paragraph to be something like

> Segregating witness data naturally provides the opportunity to replace the P2SH payment format and update existing opcodes, because there are no prior constraints on interpretation of segregated witness data; and because only the new method of doing signatures is available for witness data, the existing limits on old signatures still apply. In contrast, doing these fixes independently would increase the complexity of the Bitcoin codebase due to the need to handle different features being active at different times on the blockchain, and additional complexity in mining and validation code would be required to maintain the limits on old signature operations.

Would be nice to say it more simply, but does the above at least seem accurate?

ajtowns: okay, fixed in 88482c3 i think

ajtowns:
Hmm, checks are failing because /en/releases/0.13.1 is dated for 2016-10-31 which hasn't happened yet :(

btcdrak:
Once again, amazing work. Thanks @aj!

PR #241, opened 2016-10-27
Add release blog post for 0.13.1

harding:
This adds a release blog post for 0.13.1.

**TODO**

- [ ] This assumes the release date is Monday 31 Oct 2016.  If it as any other date, the file needs to be moved and the YAML header needs to edited.
- [ ] At the very top of the document, a link needs to be added to @laanwj's PGP-signed release announcement (as we did with the previous release).  The tests should fail until this is added, but I see they succeeded, which I guess means I'll have to fix that...
- [ ] The SHA256 hashes of the release files should be added to the bottom of this document (the "Hashes for verification" section) as was done for the previous release.
- [x] https://github.com/bitcoin-core/bitcoincore.org/pull/240 needs to be merged before this, or this document needs to be edited to stop linking to it.  (The tests should also be failing because that's missing in this branch...)

---

A preview for the blog post as of fd373869eab2e473c7953efe3098d2d604a2f9c3 may be found at: http://imgur.com/a/gKi9W

The post entry on the main page looks like this:

Please note that if you attempt to generate a preview of this document, you will need to add `--future` to Jekyll's parameters because the post is dated for 2016-10-31.

Several people provided suggestions on this in draft form, for which I am very grateful. Any errors that appear in this version are entirely my fault.  For anyone who read the final draft version, I've made a few small changes:

1. I further shortened some of the descriptions in the new features summary section at the top.  The only significant phrasing change is in point 7 (More efficient almost-full-node security).  I added a few paragraph breaks in the corresponding detail subsections but no words were changed there.

2. To the end of the document, I added a new Conclusion section (and also the template Hashes For Verification section) to match the format of the 0.13.0 release blog.

3. I changed the opening to the null dummy section to make it flow a little better.

As requested in some of the draft feedback, I added the ability to link directly to each of the items on the timeline.  This is a little bit ugly, so I'm open to suggestions for making it look nicer.  Screenshot below (the `#` are both the link and the anchor):

PR #244, opened 2016-10-27
Travis: build all possible pages for testing

harding:
In https://github.com/bitcoin-core/bitcoincore.org/pull/241#issue-185549835 I mentioned that the Travis CI tests were passing when they were expected to fail.  This commit resolves that issue.

The problem was that pull request included a page with a future date and the previous Travis configuration didn't build or test pages with future dates.  As of this commit, Travis should build all possible pages (as best I can tell from `jekyll build --help`) so that they are all tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ cache: bundler
 ## Disable external link checking to prevent spurious failures because
 ## of other people's downtime. This can also avoid wasting their
 ## bandwidth.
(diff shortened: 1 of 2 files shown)


btcdrak:
Maybe need to update the README.md with this too?

harding:
@btcdrak good catch.  README updated in  	c1f8187

PR #245, opened 2016-10-28
Segwit costs: s/fixed/updated/

harding:
@ajtowns 

> Since segwit only allows increased signature operations via the fixed opcodes, the old opcodes remain naturally limited.

This commit makes one tiny change.  When reading this sentence, I got confused because I read "fixed" as meaning "a fixed (unchanging) point" whereas I think the desired meaning was "repaired".  I think saying "updated opcodes" works well here.


ajtowns:
Sounds good to me, ACK.

PR #253, opened 2016-11-03
add Cryptobuyer to Segregated Witness Support list

angelralf:
Please add our project to the SegWit support list! Thx


angelralf:
Thanks a lot!

PR #266, opened 2016-11-18
Add BTCC to segwit support list.

mikaelw:


btcdrak:
Your pool isnt producing segwit version flag.

jl2012:
They just produced the first segwit version block at 439531

PR #288, opened 2016-12-07
Add support from Bitonic

diederikk:


NLNicoo:
Note: it says "borker" instead of "broker"

PR #342, opened 2017-03-02
Meeting notes: 2017-02-23

harding:
Notes from last week's meeting.


MarcoFalke:
ACK with nit:  s/same exact/exact same/g

PR #345, opened 2017-03-04
Use Ciphrex for name in adoption page rather than mSIGNA.

CodeShark:
Ciphrex would like to use the company name rather than the name of one of our products in the adoption page. In addition, it should be listed as a library as well as a wallet.


luke-jr:
Maybe "Ciphrex / mSIGNA" so people recognise it?

PR #370, opened 2017-04-10
CN translation of scaling optimizations post

brg444:
Translation provided by Shao Long and reviewed by Johnson Lau.


btcdrak:
ping @jl2012

jl2012:
ACK. However, if the target is non-technical audience, better avoid jargons like "uses insights from Big-O complexity concepts" and keep thing simple. Also keep the English simple (not more than 2 clauses per sentence) for easy translation

PR #394, opened 2017-06-17
Releases: 0.14.2 release notes and announcement

harding:
Announcement is just boilerplate and some slightly rearranged text from the release notes.

Preview:


laanwj:
ACK, thanks!

PR #401, opened 2017-07-05
Use local JQuery and fonts rather than Google hosted versions

harding:
Re: #396 

Preview: http://dg0.dtrt.org/

@btcdrak Some quick testing shows that the main page for the site (one of its smallest pages) is about 270 KB with all resources and loading JQuery adds 90 KB, which seems reasonable to me but you know more about the server's bandwidth limitations.


TheBlueMatt:
We're fine on bandwidth. Moving the binaries there is gonna be orders of magnitude more than any site (though I'm actually surprised that jQuery is that large... Isn't there a "cut down" version with just what we need?).

Loading fonts locally would be nice too (as long as they're loaded last).

harding:
Added locally-hosted copies of the same fonts we used before, now loaded from the end of the page footer.  Per Google, the [font license](https://fonts.google.com/attribution) is SIL 1.1 (license text contained within this commit).  Here's what the FSF says about it ([source](https://www.gnu.org/licenses/license-list.en.html#SILOFL)):



Since `.ttf` files have occasionally been used in to attack browsers, here are the wgets I ran in case anyone wants to verify this commit is providing faithful copies of the fonts:

```
wget https://fonts.gstatic.com/s/ptserif/v8/03aPdn7fFF3H6ngCgAlQzC3USBnSvpkopQaUR-2r7iU.ttf
wget https://fonts.gstatic.com/s/ptserif/v8/QABk9IxT-LFTJ_dQzv7xpJ0EAVxt0G0biEntp43Qt6E.ttf
wget https://fonts.gstatic.com/s/ptserif/v8/EgBlzoNBIHxNPCMwXaAhYPesZW2xOQ-xsNqO47m55DA.ttf
wget https://fonts.gstatic.com/s/ptsansnarrow/v7/Q_pTky3Sc3ubRibGToTAYjnFfol0SnfBMmbnmUK0fZM.ttf
wget https://fonts.gstatic.com/s/ptsansnarrow/v7/UyYrYy3ltEffJV9QueSi4UpeNX8RPf6i6WQfJWyCWEs.ttf
```

The preview is updated: http://dg0.dtrt.org

harding:
@TheBlueMatt regarding JQuery size, it looks like the minimal webserver bundled with Jekyll doesn't support gzip, so I was seeing the full file sizes.  My preview server dg0.dtrt.org does support gzip, and it shows JQuery as 32KB, which is the same size reported on the JQuery website.

TheBlueMatt:
I believe gzip is turned off on bitcoincore.org, though it's likely not useful to do so. We can turn it back on if it's worth a lot.

btcdrak:
@TheBlueMatt please do re-enable gzip compression. That's a basic staple. It's not about your bandwidth costs, it's about page speed load time and gzip makes a big difference.

TheBlueMatt:
Indeed, I disabled it without thinking cause it would be bad if we had any dynamic content, but we don't on the main server.... Will do it when I get a chance.

btcdrak:
utACK

btcdrak:
@harding I'll merge this tomorrow after server maintenance.

PR #426, opened 2017-08-30
add deployed flag to segwitsupport.csv

prusnak:
Also maybe it would be worthwhile to split "businesses" and "libraries" into two separate tables, as "libraries" cannot deploy their changes and "ready" is the final state for them.


prusnak:
If I am wrong that only TREZOR and Ledger deployed this, write it in the comment and I'll amend my commit.

AdamISZ:
@prusnak Could you put 'ready' or 'deployed' for joinmarket, it's been in use since Aug 25th. (we are running concurrently in the same pit with and without segwit, with two different code repositories, but whatever).

btcdrak:
Thinking on this we will need to update the status counter at the foot of the table with the new state.

prusnak:
Commit updated to include the requested changes.

Also I would vote to remove all "planned" entries from the table. There was plenty of time to at least start working on Segwit and switch to "wip"

AdamISZ:
> Commit updated to include the requested changes.

Joinmarket wasn't changed?

If it matters, the update is at https://github.com/Joinmarket-Org/joinmarket-clientserver

prusnak:
is Joinmarket a library or a service? If it is library, I will update it to read, if it is a service, I'd like to see it working (provide URL), and will update it to deployed.

AdamISZ:
Joinmarket is p2p coinjoin software; the only one currently in operation (been like that for the last couple of years). The main repo is at https://github.com/Joinmaket-Org/joinmarket (it may give you a more realistic sense of participation, since the repo i linked is only the new implementation with segwit).

I'm surprised you've never heard of it, but oh well. There is no service to point you at, but you can view the public offers at https://joinmarket.me/ob/ if it matters. In any case, please either update the entry or remove it, since segwit (bip49 based) is now provided as i said above.

prusnak:
Ok, so you are a library not a service. Will change the field to "ready".

AdamISZ:
Thanks, but it's not really a library. It's an application that people run. If it were me I would put "deployed". Well, no matter.

PR #427, opened 2017-08-30
joinmarket segwitsupport update

AdamISZ:
A segwit version of Joinmarket is deployed


AdamISZ:
sorry doh, half asleep

PR #431, opened 2017-09-06
Note linked project has unpatched vulnerabilities

ysangkok:


gmaxwell:
This is silly. Libsnark necessarily actually achieve the advertised zero knowledge, sure-- but that has little to do with a simple demo.  It isn't like anyone is actually going to be buying sudoko solutions.

NAK.

PR #457, opened 2017-10-13
Add japanese translation for download page

azuchi:


harding:
@azuchi Thank you for translating!  Based on your experience, is there anything you'd suggest changing in order to make translation easier or better for other translators in the future?

azuchi:
@harding Thank you for review! For translation,  [CONTRIBUTING.md](https://github.com/bitcoin-core/bitcoincore.org/blob/master/CONTRIBUTING.md) is written to use Transifex,  but I do not know it is working.

btcdrak:
Unfortunately Transifex made some changes that basically broke our workflow so that kind of fell apart.

azuchi:
Oh, I see. So it may be better to update CONTRIBUTING.md according to the current. Some people are requesting to join in Transifex.

PR #460, opened 2017-10-21
Download page: change "source code" link to current tag

harding:
Requested by @MarcoFalke in https://github.com/bitcoin-core/bitcoincore.org/pull/452#pullrequestreview-70759560

I confirmed using a local preview that the current link points to https://github.com/bitcoin/bitcoin/releases/tag/v0.15.0.1

Not implemented is @MarcoFalke's suggestion to change the text associated with the link.  I think "source code" is more understandable to most people than "git source".  If there is still a desire to change the text, I'm happy to do that.

### DIFF
--- a/_includes/templates/download.html
+++ b/_includes/templates/download.html
@@ -69,7 +69,7 @@
       <a href="{{ PATH_PREFIX }}/SHA256SUMS.asc" class="dl">{{ page.downloadsig }}</a><br>
       <a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX }}.torrent" class="dl">{{ page.downloadtorrent }}</a>
         {% if magnet %} <a href="{{ magnet | replace: '&', '\&amp;'}}" class="magnetlink" data-proofer-ignore></a>{% endif %}<br>
-      <a href="https://github.com/bitcoin/bitcoin" class="dl">{{ page.source }}</a><br>
+      <a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX}}.tar.gz" class="dl">{{ page.source }}</a><br>
       <a href="/en/releases">{{ page.versionhistory }}</a>
     </p>
     <p class="downloadkeys">


jonasschnelli:
Seems logic to link to the source code of that particular release... but why not to the tar gz file in the current release download folder (link https://bitcoincore.org/bin/bitcoin-core-0.15.0.1/)?
Linking to GitHub instead staying on the site where the sources are easy verifiable and served under our domain seems preferable.
There could also be a link that links to GitHub master ("newest unstable source –code" or similar)...

_includes/templates/download.html
```diff
@@ -69,7 +69,7 @@
       <a href="{{ PATH_PREFIX }}/SHA256SUMS.asc" class="dl">{{ page.downloadsig }}</a><br>
       <a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX }}.torrent" class="dl">{{ page.downloadtorrent }}</a>
         {% if magnet %} <a href="{{ magnet | replace: '&', '\&amp;'}}" class="magnetlink" data-proofer-ignore></a>{% endif %}<br>
-      <a href="https://github.com/bitcoin/bitcoin" class="dl">{{ page.source }}</a><br>
+      <a href="https://github.com/bitcoin/bitcoin/releases/tag/v{{CURRENT_RELEASE}}" class="dl">{{ page.source }}</a><br>
```
MarcoFalke: The archive generated by GitHub might not be deterministically created. I'd prefer to link to the archive in our /bin/ folder. (And then optionally to the general Bitcoin Core GitHub, as well)

harding:
@jonasschnelli @MarcoFalke Thanks for the feedback; I misunderstood the original request.

I changed the link to the tarball in /bin.  I manually tested that it points to the correct relative URL, and also tested that the test script run by Travis ensures the tarball exists on the live server (which is mainly intended to catch situations where the file naming pattern is changed).

I didn't add a link to the GitHub repo as the page-top navigation has a "development" menu with those details and I think this section of the download page is getting a bit crowded (but I'm open to putting a link here anyway if you want).

MarcoFalke:
utACK 86530dfbb2d0892ebeb369bad5a7e33691d86da5

PR #475, opened 2017-12-14
Update segwitsupport.csv

mstrokin:
Added CoinMall marketplace


mstrokin:
image
work for both BTC and LTC

PR #479, opened 2017-12-19
Note that aggsig paper was rejected

ysangkok:


harding:
Thanks.  You link to a libsecp256k1 PR as a source for the Bellare-Neven paper being implemented, but I don't see any reference to that in the PR comments, the commit message, or the diff.  I haven't been following the Schnorr implementation discussion, so I'm not in a position to doubt this assertion, but it would be nice to have some confirmation that the algorithm being implemented is the one from the paper.

apoelstra:
That PR is indeed for Bellare-Neven. It is not API-complete and especially not doccomment-complete, hence the missing reference.

ysangkok:
@harding I have changed the tense and added a link to the CoreDev Event website.

harding:
@ysangkok Looks great, thanks!

jnewbery:
I don't think that old blog posts should be updated to be kept current. This isn't supposed to be living documentation.

Work on Schnorr signatures is ongoing, and we're likely to see lots of activity in 2018. We shouldn't try to keep this old blog post up to date as Schnorr signature, signature aggregation and batch validation develops.

So weak concept NACK from me, but I won't stand in the way if other people think this is worthwhile.

ysangkok:
I see great value in having current roadmaps. Maybe this just shouldn't be a blog post.

harding:
This PR hasn't gained any traction in several months, so I'm going to close it (but it can be reopened if there's disagreement).  If I could go back in time, I'd suggest that we shouldn't have published the blog post about signature aggregation on BitcoinCore.org, as potential future directions for Bitcoin Core and the Bitcoin protocol should probably be posted on the websites of the people directly working on those proposals, as Blockstream did for their subsequent post about [MuSig signature aggregation](https://blockstream.com/2018/01/23/musig-key-aggregation-schnorr-signatures.html).

@ysangkok thanks again for working on this.  I'm sorry this isn't the outcome you wanted.

PR #480, opened 2017-12-19
Upgrade site build dependencies

harding:
The commits in this PR upgrade the site's Jekyll-based build system to the latest versions of the available software.  This results in one notable change to the site content:

- Automatically-created reference ids (commonly called "slugs" in webdev) now include numbers whereas they didn't before.  For example, previously we had a subheading "0.13.0 Change Log" with the slug `#change-log` that is now `0130-change-log`.  Unfortunately, this change may break some links to the site (in the commits, I fixed all broken internal links); however, any broken links should still point to the correct page---they just won't automatically scroll to the intended section.

Except for the above changed links and some typo fixes described in the commit messages, the site should be identical to before.  Here is a preview for anyone who wants to spot check pages: http://dg0.dtrt.org/

## Verification

Since we haven't upgraded the software in almost two years, there are a lot of dependencies changes.  Anyone who wants to check that I'm not doping the `Gemfile.lock` file can perform the upgrade process themselves, which is quite simple:

1. Switch to a commit before the "Build: update dependencies" commit
2. Run: `bundle update` in the top-level source directory
3. Compare your generated file against my file in the above commit.  Note: I don't think the ordering of the dependencies is deterministic, so you may need to sort the two files.

## Upgrade instructions

Anyone who has previously built local copies of the site will need to update their dependencies by running the following command in the top-level source directory: `bundle install`

The dependencies will also need to be updated on the build server when this commit is merged, so this should probably only be merged by someone with local/remote access to the system console.

---

The rationale for all changes can be found in the message for the encompassing commit.


jonasschnelli:
Can someone verify this PR: @MarcoFalke, @jnewbery?

MarcoFalke:
Needs rebase

harding:
Rebased and also pushed an updated preview to my preview domain.

MarcoFalke:
I think for testing it could make sense to check in the html into a fresh git (before the upgrade and after) and then run `git diff`

harding:
@MarcoFalke yeah, I did skim the diff for my own testing, but it's pretty huge mostly because of minor changes in the output of the upgraded syntax highlighter.  However, you can find an HTMLized diff here: http://dg0.dtrt.org/diff.html

MarcoFalke:
utACK 6fad7872de3b05e872d57d09c42bad35410c4895

MarcoFalke:
Needs rebase

harding:
Closing as this doesn't seem to be getting the requested testing and it's kind of a pain to rebase due to the need to skim a huge HTML diff.  The Jekyll project is [planning](https://jekyllrb.com/news/2018/04/19/development-update/) a new major version for September that will have breaking changes for this site (but also maybe some nice new features), so I'll aim to try another update after that's out.

PR #484, opened 2017-12-21
Add Bitflyer No Support

jefferson-1:


jonasschnelli:
Is there a public announcement or statement that they won't support SW?

jefferson-1:
Well...this now needs to be changed. They support Segwit on sending but not receiving. So sending to their exchange requires full fee. https://www.reddit.com/r/Bitcoin/comments/76rpu3/bitflyer_adds_segwit_support_to_withdrawals/

FndNur1Labs:
they build own engine

PR #518, opened 2018-02-28
rm segwit support

MarcoFalke:
A plain collection of links with no moderation or disclaimers only serves as a billboard for free advertisement.


jonasschnelli:
Yes. Agree that we should remove it at some point.

jnewbery:
Concept ACK. This list generates a huge amount of churn. I personally consider Segwit to be the default, normal way to use Bitcoin. No need to list companies that are using Bitcoin in a standard way!

harding:
Concept ACK.

My only criticism is that I think we should avoid removing long-standing pages without leaving a redirect to an alternative resource.  Happily, we already have a redirect plugin enabled, so that's pretty easy:

```diff
diff --git a/_posts/en/pages/2016-10-26-segwit-upgrade-guide.md b/_posts/en/pages/2016-10-26-segwit-upgrade-guide.md
index 303896a..eaa53b5 100644
--- a/_posts/en/pages/2016-10-26-segwit-upgrade-guide.md
+++ b/_posts/en/pages/2016-10-26-segwit-upgrade-guide.md
@@ -8,6 +8,9 @@ lang: en
 permalink: /en/2016/10/27/segwit-upgrade-guide/
 version: 1
 excerpt: Almost two years of iterative design, development, and testing has gone into the version of segwit being released in Bitcoin Core 0.13.1, with much of the effort over the last year focused on making it as easy as possible for existing Bitcoin users, businesses, developers, and miners to upgrade to segwit.
+redirect_from:
+  - /en/segwit_adoption/
+  - /zh_CN/segwit_adoption/
 ---
 {% include _toc.html %}
 {% include _references.md %}
```

You can see a preview of what the redirect looks like by visiting either of the following preview URLs:

- http://dg0.dtrt.org/en/segwit_adoption/
- http://dg0.dtrt.org/zh_CN/segwit_adoption/

The segwit dev guide seems to me like the best page to redirect to since I think the adoption page was originally intended to spur developer implementation of segwit, but the [segwit benefits](https://bitcoincore.org/en/2016/01/26/segwit-benefits/) page might be preferred by others.  Either page (or pretty much any other page on the site) is fine by me.

achow101:
Concept ACK

btcdrak:
This page was intended as a tracker for segwit activation and adoption in the early days. It's clearly not required anymore. Concept ACK.

jnewbery:
> I think we should avoid removing long-standing pages without leaving a redirect to an alternative resource

ACK @harding's change

@MarcoFalke - thoughts?

karel-3d:
Only half related to this PR, but it would make sense to have a list of businesses supporting *native segwit*,

Most of Bitcoin services do not support sending to bech32 native segwit addresses. Some don't even support sending to their addresses *from* bech32 (happened to me with coinbase payment gate).

MarcoFalke:
Added redirect

jnewbery:
Tested ACk def532dabc92ca010d6f1abb60ed5b26c35b2c2f

MarcoFalke:
Anything left to do here?

MarcoFalke:
@btcdrak @jonasschnelli ^

Issue #470, opened 2017-11-24
Update the team-page


jonasschnelli:
The team page should be updated (add/remove some devs).

I propose we add...
* @morcos
* @jnewbery 
* @ryanofsky
* @promag
* @achow101 
(more?)

Maybe remove...
@pstratem (seems no longer active)

MarcoFalke:
Ack

promag:
ACK I guess :P

jnewbery:
I'm personally not a fan of the team page, and would prefer to see it removed. It reinforces the mistaken belief that 'Bitcoin Core' is a group/organization.

I think it's ok (although not necessary) to have the maintainers listed, since they have a functional role. I don't think it's a good idea to include 'contributors'. There's no difference between me and any random passing contributor - we each have the same rights to open PRs and comment on open PRs.

I'd prefer to see the **contributors** section of the team page reduced to something like:

> Everyone is free to propose code changes and to test, review and comment on open Pull Requests. Anyone who contributes code, review, test, translation or documentation to the Bitcoin Core project is considered a contributor. The release notes for each Bitcoin Core software release contain a credits section to recognize all those who have contributed to the project over the previous release cycle.

(If we do decide to keep the team contributors section, no need to include me. Thanks!)

MarcoFalke:
I think the motivation for the team page was to convey that there are many who contributed commits and then list the ones that had the most commits in the previous releases. Probably to give them additional credit (as not everyone digs up the release notes).

Though, over the months, no one seems to update that page and it might convey a false sense of "closed group membership".
I think adding the paragraph that @jnewbery suggested and linking to https://github.com/bitcoin/bitcoin/graphs/contributors?from=${12_MONTHS_AGO}&type=c gives all of the recent contributors some additional credit beside the release notes without having to update the page or create a false impression of a fixed group.

jnewbery:
> I think adding the paragraph that @jnewbery suggested and linking to https://github.com/bitcoin/bitcoin/graphs/contributors?from=${12_MONTHS_AGO}&type=c gives all of the recent contributors some additional credit

Yes, I think that's fine, although wording should be careful - number of commits is not a good metric for how much an individual has contributed (it doesn't capture maintenance tasks, review, test, documentation, etc)

morcos:
[...]

PR #519, opened 2018-02-28
rm list of contributors

MarcoFalke:
The list was severely outdated and could potentially create the impression that there is a closed set of contributors to the the open source software client.


MarcoFalke:
See discussion in  #470

_data/contributors.yml
```diff
@@ -23,116 +23,11 @@ maintainers:
...
-    role:
-  -
-    name: paveljanik
-    github: paveljanik
-    irc: paveljanik
-    twitter:
-    avatar: https://avatars1.githubusercontent.com/u/6848764?v=3&s=460
-    bio:
-    role:
+    role: Wallet maintainer
```
jnewbery: 😕

MarcoFalke:
@jnewbery I am not objected to a replacement of the team page that makes more sense. However, I think first removing the current team page is orthogonal to that and an improvement on its own.

(Also amended the section to what you suggested in the comment you just linked to)

jnewbery:
> orthogonal

I don't think this word means what you think it means :)

jnewbery:
utACK c8618b1359005d78b584c95b73d3856a84c52736

luke-jr:
This seems to only get rid of some of the team, and not others, giving an erroneous impression that merge permissions is more important than it really is. IMO either the page should be removed entirely, or otherwise updated to a current list. NACK removing some active people and leaving others...

MarcoFalke:
Done

luke-jr:
Okay, concept ACK then I guess. (Not really setup to render/test it myself.)

jnewbery:
> giving an erroneous impression that merge permissions is more important than it really is

Not really - the difference is that a maintainer has a well-defined role in the github organization. They have permissions to merge code. I don't think identifying them infers that they're any more or less 'important'.

Concept ACK removing all, or just leaving maintainers.

jl2012:
concept ACK remove all

harding:
Concept ACK for remove all, but I suggest we leave a redirect to another page on the site like the [contribute page](https://bitcoincore.org/en/contribute/) so the URL isn't broken.  For instructions on creating the redirect, see https://github.com/bitcoin-core/bitcoincore.org/pull/518#issuecomment-369680701

achow101:
Concept ACK

laanwj:
I kind of like the idea of giving frequent contributors a more visible place then at the bottom of release notes, if they want. But I also agree that if it doesn't get updated it's better to remove it.
Also we already have the team pages on the bitcoin / bitcoin-core organizations, and keeping these things updated in parallel is a pain.

MarcoFalke:
@harding Can you elaborate a bit on that. I think I only removed files in the `_data` and `_include` directories and only reformulated the actual `_post`. Not sure which page needs a redirect...

laanwj:
I think he means that the URL should lead to a redirect in case external sites might be pointing as that's clearer than showing a not-found page.

MarcoFalke:
What does "the URL" refer to?

On Mar 3, 2018 13:46, "Wladimir J. van der Laan" <notifications@github.com>
wrote:

I think he means that the URL should lead to a redirect in case external
sites might be pointing as that's clearer than showing a not-found page.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/bitcoin-core/bitcoincore.org/pull/519#issuecomment-370170400>,
or mute the thread

harding:
@MarcoFalke sorry, I thought this PR removed the team page but I just confirmed that it only removes the list of contributors from that page.

Tested ACK 41bd570ae68fb8831a584eb816df371d2236e99f

Preview of the updated page:

PR #520, opened 2018-03-06
A great website is not in the list!!

shivamarkss:
Hello Guys,

I saw some website here and which is similar as this website. I'm a betting from a long period and I really liked that game.


randolf:
@shivamarkss You've mentioned that the web site you want added is "great," but you haven't indicated whether the web site supports SegWit (Segregated Witness) technology.  Since the file you modified is a list of web sites that supports SegWit, I suggest that you find out whether this is the case then report your findings (e.g., such as an eMail reply from the web site's operator{s}).  Thanks.

Issue #525, opened 2018-03-09
Add documentation about generated javascript


karel-3d:
Jekyll does not generate the minified javascript; it was generated in the initial commit and the process is not documented.

(It is questionable if the javascript even needs to be minified, since the size difference is minimal in this case.)

harding:
I agree that we should have a non-minified copy of the JS in the repository for review (if necessary for optimization, we can minify it at compile time).  I believe the origin of the version we use is from an older version of the [Minimal Mistakes](https://mmistakes.github.io/minimal-mistakes/) Jekyll theme in case anyone wanted to look into it.  (If not, I'll investigate myself when I get a chance.)

PR #526, opened 2018-03-10
Adding RPC documentation

karel-3d:
I have added auto-generated RPC docs.

I think the best place for Bitcoin Core documentation is Bitcoin Core website.

The script is in doc-gen folder.

I have reused ToC component for the RPC list; this needed some change in the javascript. I unfortunately needed to edit the minified javascript by hand (see https://github.com/bitcoin-core/bitcoincore.org/issues/525)

It looks fairly good in my opinion.


saleemrashid:
I think a cleaner method would be to generate Jekyll data files, then use Liquid templating for the actual documentation.

saleemrashid:
Additionally, since the script is quite simple, it might be preferable to rewrite it in Ruby to avoid extra dependencies.

karel-3d:
Hello @saleemrashid , thanks for feedback 

> I think a cleaner method would be to generate Jekyll data files, then use Liquid templating for the actual documentation.

Would it be that much different from the current version? Right now, the generated files are just the documentation wrapped around with `<pre>`, plus one include. With what you propose, the result would be without the `<pre>` and the `<pre>` would be in the liquid template, right.

It would be probably simpler I guess... I can do that

> Additionally, since the script is quite simple, it might be preferable to rewrite it in Ruby to avoid extra dependencies.

I guess? I am not learning ruby for this PR :), also the script needs to be run just once after release, by anyone.

I am not rewriting the script to Ruby.

karel-3d:
I am looking at this, it seems like a better fit than data files

https://jekyllrb.com/docs/collections/

karel-3d:
Should I continue working on this, is there an interest?

jonasschnelli:
Thanks for adding this @karel-3d!
Agree with @saleemrashid for the auto-generation approach.
Concept ACK

karel-3d:
cool, I will re-work this as Saleem suggested

karel-3d:
I have reworked the docs as a collection, since I think it's better suited for that than data files (or pages/posts). The navigation table on the right is now auto-generated by jekyll, based on `btcversion` and `btcgroup` attributes, as is the menu and partly the indexes.

I am still not sure if I want to rewrite the script to Ruby. I would very much rather not. :( 

@saleemrashid commented on old version of the PR

https://github.com/karel-3d/bitcoincore.org/commit/c0fa17f91066d7f9c4d1676f1ddaefe363cd12e7#r28709165

I will reply here - I did not want to use stuff like `.sibling` since that would too tightly couple the HTML layout to the javascript. I don't particularly care though.

saleemrashid:
@karel-3d JavaScript to this effect would also work:

```javascript
$(".toc").each(function(){
  var drawer = $(this).find(".drawer");
  $(this).find(".toc-header").click(function(){
    drawer.toggleClass("js-hidden");
  });
});
```

karel-3d:
I am still not sure what to do with the javascript (related to https://github.com/bitcoin-core/bitcoincore.org/pull/552 ). 

Currently, there is  assets/js/_main.js , which is actually irrelevant, and assets/js/scripts.min,js. As far as I can tell, the _main.js is not used anywhere, and the scripts.min.js was once generated from it, but is now static.

In the PR, I made the change in both manually. But that feels wrong :)

harding:
If you want to simply the whole JS setup and skip minifying them if it's not significantly affecting size, I'd welcome that (perhaps as a separate PR that you rebase this upon).  Perhaps a good way to do that would be to put each source JS file in an `_include` folder and then make assets/js/scripts.js something like:

```text
{% include js/foo.js %}
{% include js/bar.js %}
```

Looking at a load of the current main page, I see we already have three separate JS file getting loaded (scripts, jquery, and modenizr) that could be combined to eliminate some TCP and HTTP overhead.

In addition, such a PR should be easy to review since it'd just be moving files from the `assets` directory to the `_includes` directory and then creating a file that concatenates them on site compilation.  It wouldn't be challenging like checking to see that no additional JS was injected into a minified JS file.

If we ever wanted to minify them, we could later write a plugin to pipe that content to a minifier, e.g.:

```text
{% minify %}
  {% include js/foo.js %}
  {% include js/bar.js %}
{% endminify %}
```

(Note: I'm not actually sure of the order of operations for Jekyll/Liquid tags, so maybe that wouldn't work.  Still we could probably cobble something together.)

karel-3d:
Now waiting for #554 , will rebase after that is ACKed

karel-3d:
I have rebased the PR on top of #554 

@saleemrashid I have fixed the JS to use your code, thanks, it is probably cleaner

@harding 

Are you sure about the missing URL on mobile? I can see it.

It is, however, true, that the "RPC Doc" menu does not further expand into submenu for all the versions. I think it's OK - clicking on the link will get you to the general RPC index, which has the versions listed.

It seems like a bug in the template for sub-menus - see About-Sponsorship on the current website. It does not expand (because you need to click for expansion, and click leads to the index).

karel-3d:
Ad 6 rows in the menu:  Oh, that makes sense. (Hunting bugs in the jekyll theme is hard :))

Ad other previous comments:

* I have added some instructions to the script itself
* I have moved the script to contrib
* The files were escaped by the golang default templating engine, "html/tempate"; however, I added the highlight thing and that made the script even simpler (and easier to rewrite in other language :))
  * However, that produced a weird bug when the highlight "overlapped" the top menu. I corrected that in the commit "Corrected strange CSS bug...". I still have no idea what really causes the bug, but CSS is hard.
* Both the directories and the URLs are renamed, to `/en/doc/0.16.0/rpc/` (so the REST API can also be versioned).

... I think that's all so far :D

karel-3d:
good point on 1., I will add another commit that fixes that.

2. is a stylistic choice, I won't do that for now

karel-3d:
...done

harding:
Tested ACK for updated tip 7625ce854cf90d8651c1ec611e9a981196961ae4

I'm completely satisfied with the code in this PR and with the ease of maintaining it through future releases, but I would appreciate one or two additional quick reviews of the things that will be annoying to change after the merge.  If particular:

- The URL scheme, as once people start linking to stuff, I don't want to break those links by moving stuff
- The amount of information provided.  We can always add more information later, but if we're currently displaying too much (e.g. regtest-only RPCs like `generate`), removing that later may break user expectations
- Whether there's any information that should be present from the start but which isn't in this PR, like maybe warnings about sanitizing results before using them in sensitive contexts, etc...

A preview is available here: http://dg4.dtrt.org/en/doc/0.16.0/

saleemrashid:
I think the only decision to be made about the URL is whether `doc/<version>/rpc` or `doc/rpc/<version>` is preferable.

harding:
@saleemrashid between those alternatives, I personally the current `/doc/<version>/rpc/<rpc-name>` for the reason @karel-3d gave: it makes it easy to add docs for (e.g.) the REST interface at `/doc/<version>/rest/<endpoint>/` and then change `/doc/<version>` to link to both `./rpc` and `./rest`.  Also, I have an ongoing slow project to write some other user documentation that I'd like to be version-specific that I think would probably nicely fit into that schema, e.g. `/doc/<version>/gui/backup`

An additional nice advantage of the above that @karel-3d discussed out of band would be that one could more easily create diffs between versions to get a summary of all the documentation changes.  E.g.,

```bash
diff -ru '_site/doc/<old_version>' '_site/doc/<new_version>'
```

Do you think `/doc/rpc/<version>/<rpc-name>` would be better?  I think it might produce a better "fuzzy matches" for visitors who are looking for basic documentation and don't care if it exactly matches their program version.  For example, imagine we've uploaded RPC docs for 0.15 and 0.16 but REST docs for only 0.15.  Alice is running 0.16 and wants docs on the `GET /rest/tx` endpoint; her navigation through the site starting from `/doc` may look like one of the paths below:

- Current `/doc/version/category/name`: /doc -> /doc/0.16 -> END, no rest docs found
- Proposed `/doc/category/version/name`: /doc -> /doc/rest -> /doc/rest/0.15, she knows 0.16 isn't available but 0.15 docs are probably good enough

I think I still favor the current scheme, but maybe there other reasons to favor the alternative?

karel-3d:
I used `/doc/<version>/rpc/<rpc-name>` because it will make it easier to later add different type of docs, as you mentioned.

I however don't have a really that strong preference for it. I don't think Alice will really have problem with either scheme.

saleemrashid:
I think that the current scheme is preferable, and a `/doc/<category>` can be added at a later date (containing a list of versions, linking to `/doc/<version>/<category>`

karel-3d:
I can add 0.16.1 docs if it's released before then :)

contrib/doc-gen/generate.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"text/template"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+const BITCOIN_PATH = "/home/g/dev/bitcoind/bitcoin-0.16.0/bin/bitcoin-cli"
```
saleemrashid: This should be changed. Since `bitcoin-cli` versions is generally forwards- and backwards-compatible, using whichever version is in `$PATH` should be fine.

contrib/doc-gen/generate.go
```diff
@@ -0,0 +1,152 @@
...
+const VERSION = "0.16.0"
```
saleemrashid: Can't this be obtained with `getnetworkinfo` rather than hardcoded?

contrib/doc-gen/generate.go
karel-3d: Yes

contrib/doc-gen/generate.go
karel-3d: Yes

karel-3d:
@saleemrashid has objections to the script - which was frankly a bit cobbled together (paths hardcoded etc). I will try to fix those, but not sure if on monday.

harding:
Cool, thanks.  Provided that the script changes aren't complicated, I'll probably review them when ready and merge soon after, as long as it isn't too near a release.

Also, FYI for anyone who cares, I tested the commits in this PR rebased on top of #560 and they passed the repaired test.  (There's no need to rebase the actual PR; it should merge cleanly as-is.)

karel-3d:
The script is cleaned up a bit

harding:
Re-tested re-ACK 31f77ada0859e88dcd1ae71c5e7ec02c384c7e9f  Thanks!

karel-3d:
And again, thanks @saleemrashid :)

karel-3d:
Thanks for the merge and publishing! :)

As I said on Twitter now, the formatting is not ideal, but I don't know where to even begin with that...

screenshot from 2018-06-12 19-31-22

Either somehow edit bitcoin core code itself and add some metadata, or parse it heuristically and hope for the best

saleemrashid:
Due to the structured way in which the RPC help strings are generated in Bitcoin Core, it could eventually be modified to build YAML files for the documentation that can be dropped in this repository.

harding:
Yeah.  I think two options would be,

1. Writing a linter run from Travis CI that enforces some structure on the documentation, so that we can parse it better.  The linter can start out quite liberal and be made more strict over time.

2. Writing formatting functions to help to standardize the output, like the example lines at the bottom of most RPCs that @saleemrashid mentions.  These functions could also start naive enough to produce identical output (hopefully making merge easier) and be made more strict about their inputs over time.  Then either through parsing of the output, patching of the functions, or maybe passing an argument to `help` we could get well-defined output out of bitcoind.

The two approaches aren't incompatible.  I think the stricter a lint checker gets, the more appeal there will be in functions that do some of the work automatically.

MarcoFalke:
I believe we do some primitive parsing internally: E.g. we assume the first line starts with the rpc name and then has all the parameters and options.

You could start doing some simple formatting based on that. Like show the rpc name bold...

Other than that, I welcome any linters that enforce structure.

mody369:
> Thanks for the merge and publishing! :)
> 
> As I said on Twitter now, the formatting is not ideal, but I don't know where to even begin with that...
> 
> screenshot from 2018-06-12 19-31-22
> 
> Either somehow edit bitcoin core code itself and add some metadata, or parse it heuristically and hope for the best
whats your offewr anmd plan and start to fucking do it and lets work

PR #529, opened 2018-03-20
Coinomi added SegWit support

erasmospunk:
SegWit support was added in the latest [beta](https://play.google.com/apps/testing/com.coinomi.wallet) of the wallet.


erasmospunk:
Update: SegWit is added in production.

PR #547, opened 2018-04-28
Readme: fix and extend instructions to install Ruby

### LINKED ISSUE #534: Incorrect instructions on README.MD
Building states :

"To build locally, install Ruby 2.2.2 or greater and then run the following commands"

When I run "bundle install" : 

"Your Ruby version is 2.5.1, but your Gemfile specified 2.2.2"

harding:
Closes #534 

- Fixes Ruby version error (previous instructions say "or greater" but an exact version match is required).
- Clarifies that `bundle` needs to be run in the source directory
- Updates the `htmlproof` command to match the syntax currently used in `.travis.cfg` ; this is necessary because the old syntax will fail on links to the `bin/` directory since that's not part of the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,20 @@ permalink: /en/2016/01/01/short-title
 
 ## Building
 
-This website is based on Jekyll. To build locally, [install Ruby 2.2.2](https://gorails.com/setup) or greater
-and then run the following commands:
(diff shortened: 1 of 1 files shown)


jnewbery:
ACK 94f550f5afbd43435343bf479149edc3d593a085

PR #550, opened 2018-05-02
Contact & Announcement List pages: update GPG key links & provide fingerprints

### LINKED ISSUE #543: dead links for gpg-keys on the announcements page (again)
It seems https://github.com/bitcoin-core/bitcoincore.org/issues/170 has reemerged.

The links on https://bitcoincore.org/en/list/announcements/join/ to the three listed keys all respond with 404. For example https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-keys/laanwj-key.pgp

harding:
Closes #543 ; CC: @kallerosenbaum

Updated bottom section on Announcement list page looks like:

---

New bottom section on updated Contact page looks like:

---

Translations of the two pages don't have any text changed, but the links for the three keys are redirected to the pgp.mit.edu keyserver web interface as suggested in the corresponding issues (pgp.mit.edu is slow, but at least it doesn't 404).  When those pages are re-translated with the updated content, we can remove the old links directly to the keys.

I obtained the key fingerprints from [here](https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-keys/keys.txt) but haven't otherwise validated them, so somebody who has checked them should probably review this PR.

### DIFF
--- a/_includes/_references.md
+++ b/_includes/_references.md
@@ -112,9 +112,9 @@
 [slack-invite]: https://slack.bitcoincore.org/
 [software life cycle]: /en/lifecycle
 
-[laanwj-key]: https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-keys/laanwj-key.pgp
-[jonasschnelli-key]: https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-keys/jonasschnelli-key.pgp
-[sipa-key]: https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-keys/sipa-key.pgp
+[laanwj-key]: https://pgp.mit.edu/pks/lookup?op=get&search=0x71A3B16735405025D447E8F274810B012346C9A6
+[jonasschnelli-key]: https://pgp.mit.edu/pks/lookup?op=get&search=0x32EE5C4C3FA15CCADB46ABE529D4BCB6416F53EC
+[sipa-key]: https://pgp.mit.edu/pks/lookup?op=get&search=0x133EAC179436F14A5CF1B794860FEB804E669320
 
 [recent-contributors]: https://github.com/bitcoin/bitcoin/graphs/contributors?from=2017-03-01
 
--- a/_includes/dev-keys.md
+++ b/_includes/dev-keys.md
@@ -1,0 +1,21 @@
+{% capture /dev/null %}
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!!! NB: for this file to render correctly, !!!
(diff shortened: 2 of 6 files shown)


kallerosenbaum:
That's a huge improvement. It even looks good!

jnewbery:
ACK 54f95ce852e19b09b00a9041ca651f1382af6340

laanwj:
> I obtained the key fingerprints from here but haven't otherwise validated them, so somebody who has checked them should probably review this PR.

Fingerprints are correct:
```
      Key fingerprint = 71A3 B167 3540 5025 D447  E8F2 7481 0B01 2346 C9A6
uid                   [ultimate] Wladimir J. van der Laan <laanwj@visucore.com>
uid                   [ultimate] Wladimir J. van der Laan <laanwj@gmail.com>

      Key fingerprint = 32EE 5C4C 3FA1 5CCA DB46  ABE5 29D4 BCB6 416F 53EC
uid                   [  full  ] Jonas Schnelli <dev@jonasschnelli.ch>

      Key fingerprint = 133E AC17 9436 F14A 5CF1  B794 860F EB80 4E66 9320
uid                   [  full  ] Pieter Wuille <pieter.wuille@gmail.com>
```
A small nit - there's something to be said for formatting the key fingerprints as above^^ per four hex digits, as this makes it slightly easier to compare.

harding:
@laanwj Yeah, I think that could help.  I just tested and GPG allows you to --recv-keys even with spaces in them, so it doesn't make the key import instructions much more complicated (people just need to add quotes).  Here's an updated preview with the latest commit:



@kallerosenbaum do you foresee any UX issues with this updated styling?

kallerosenbaum:
@harding No, not really. Since you can quote the space separated text in gpg, it's fine. You *could* add a copy button that copies the text without spaces to clipboard, but that's possibly exploitable (eg display one thing and copy another thing). Also, there might even be a way to css format the fingerprint with spaces, but the text is actually without spaces. That's possibly confusing to the user who expects a space separated fingerprint in her paste buffer because that's what she thought she copied.

A drawback with spaces is that you can't double-click to copy the whole fingerprint.

Your extra comment about quotes is clear. I think it's better with spaces but not in every aspect.

harding:
@kallerosenbaum yeah, I do think we lose a little something by not being able to double-click to select the whole fingerprint, but hopefully that doesn't end up being too much of an issue (if it is, we can always change back or experiment further).  Thanks for reviewing!

If anyone has further comments, I'd be happy to hear them.  If not, I plan to merge sometime after 16:00 UTC Friday.

MarcoFalke:
Post merge ACK by looking at the live website.

PR #552, opened 2018-05-07
Readme: mention theme used and link to its documentation

### LINKED ISSUE #525: Add documentation about generated javascript

harding:
Closes #525 ; CC: @karel-3d 

Links to the website and documentation for the upstream of the theme we use.  We haven't upgraded the files related to the theme since the website went live in Jan 2016, so I also mention that the current documentation might not completely match the theme files/settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,18 @@ pages are built and checked.
 ## Contributing
 
 Contributions welcome. Please see [CONTRIBUTING.md](/CONTRIBUTING.md) for details.
+
+## References
+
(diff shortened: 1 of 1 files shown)


jnewbery:
ACK 9865f85f7906e7aaa960df6fce1b064d8360056f

PR #554, opened 2018-05-15
Simplify javascript

karel-3d:
* Deletes unused js modules:
  * magnific-popup (tool for better image galleries) is not used at all
  * modernizr is not used at all (it creates window.Modernizr js object which is never used anywhere else)

* Deletes obsolete js modules:
  * respond and html5shiv are used only in IE9, which had last version in 2011. We can reasonably assume those will not be very frequent, and the website should work there as-is anyway.

* Puts javascripts to a single file
  * instead of using pre-built minified file, jekyll puts the files together through `{% include %}`

* Uses "local" javascript instead of hardlink to bitcoincore
  * This allows easier future experimentation

I wanted to also delete fitvids, but it is used on one page - https://bitcoincore.org/en/2016/01/13/development-visualisation-2015/ . So it is probably better to keep using it.

This PR also removes minification as a by-product of using jekyll for putting js together. I think the performance difference is negligible, since jquery is already minified.

Related: https://github.com/bitcoin-core/bitcoincore.org/pull/552 , https://github.com/bitcoin-core/bitcoincore.org/pull/526


harding:
@karel-3d Looks like one of the removed JS files was used to compact the menu on mobile.  The first image below is the live site; the second image is a preview I uploaded to dg4.dtrt.org.  These screenshots are taken using the Firefox responsive testing mode, but I can replicate on Chrome for Android.

 


I haven't found any other issues in testing with FIrefox/Linux, Chromium/Linux, or Chrome/Android.  I also used <https://netrenderer.com> to get a screenshot of the preview homepage with IE9 and it looks perfectly fine to me.

The diff was a breeze to review and looked ideal to me.  Although we'll probably have to add back in some JS to fix the mobile menu issue, as it currently stands, the new single JS file is 31% smaller than the combined size of the three previous files (in all cases after gziping and exclusive of HTTP and TCP overhead).

So, in short, this LGTM in all cases once the mobile menu issue is addressed.  Thanks!

karel-3d:
It seems to have something to do with modernizr, because the minified version seem to be not just modernizr, but some customization, which the css files rely on. I will look further.

karel-3d:
Done, see the commits. Modernizr adds `js` class to <html> root element. I have added that to main.js

karel-3d:
I might have broken a page that includes an image with a link, if such exist. That would use the magnificient-popup thing.

I have not found such a page through grepping.

harding:
@karel-3d Yeah, I saw that and also couldn't find any image-link that uses it.  If we later have some page that would benefit from a lightbox preview, we can add it back (possibly using a different JS library) and perhaps better document that the site provides such a feature (I didn't realize it before seeing your commit).

PR #559, opened 2018-06-08
Meetings index: remove excessive whitespace

harding:
A for loop that generates the list of meeting entries is adding an two additional empty lines to each entry for each additional entry, so the whitespace for *n* entries was growing at `n*(n+1)`.  With there currently being 96 entries, this was a modest 9k empty lines---not a real performance problem.  However it's ugly, it makes debugging harder, and it's clearly a bug, so I fixed it.

There's no rush to apply this commit: by my math, it'll take about 100 years of meetings before the page size could become an actual bottleneck.

Preview of the updated page (should look the same): http://dg3.dtrt.org/en/meetings/
HTML diff from before & after this commit: http://dg3.dtrt.org/diff.html

### DIFF
--- a/_includes/_meetings.html
+++ b/_includes/_meetings.html
@@ -1,20 +1,17 @@
 <ul class="meeting">
   {% assign english_posts=site.posts | where:"lang", 'en' | where:"type", 'meetings' | sort:'post.date' %}
   {% assign translated_posts=site.posts | where:"lang", page.lang | where:"type", 'meetings' %}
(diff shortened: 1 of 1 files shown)


harding:
In the process of re-testing this immediately prior to pushing a merge, I discovered a pre-existing bug in the HTML for this page that's not caught by our tests.  Since this isn't urgent, I'm going to close it for now, improve the tests, fix the bug, wait for that to get merged, and then rebase this on top of the new master.

PR #565, opened 2018-06-16
[Trivial] Download page: make expandable sections use clickable mouse cursor

harding:
Preview: http://dg1.dtrt.org/en/download/

On the live BitcoinCore.org site, if you hover your mouse over one of the expandable sections in the Download Page verification docs, it shows the default mouse cursor (mouse not pictured in image below).

With this commit, the cursor is changed to the cursor used for clicking on a link (e.g. a hand with pointer finger extended).  I've tested this works using a Chromebook running the Chrome browser (on my usual OS, nothing in any browser seems to change the cursor).

Suggested by Daniel E. Renfer, duck1123, [on Twitter](https://twitter.com/duck1123/status/1008013971830509568)

### DIFF
--- a/_sass/page.scss
+++ b/_sass/page.scss
@@ -1068,3 +1068,7 @@ div.maintainers {
 .bg-blue {
 	background-color: $bg-blue;
 }
+
+details > summary {
+        cursor: pointer;
+}


jnewbery:
utACK 7c7c25cde4ac1ee00163d200fa17bf95e1348ec9

PR #567, opened 2018-06-22
Meetings: add 2018-06-14

harding:
Preview: http://dg0.dtrt.org/en/meetings/2018/06/21/

### DIFF
--- a/_posts/en/meetings/2018-06-21-meeting.md
+++ b/_posts/en/meetings/2018-06-21-meeting.md
@@ -1,0 +1,336 @@
+---
+title: IRC meeting summary for 2018-06-21
+lang: en
+permalink: /en/meetings/2018/06/21/
+name: 2016-06-21-meeting
+layout: page
+type: meetings
+version: 1
+---
+{% include _toc.html %}
+{% include _references.md %}
+
+- View this week's log on [BotBot.me][bbm log] or [MeetBot][mb log]
+- [Meeting minutes by MeetBot][mb minutes]
+
+---
+
+Topics discussed during this weekly meeting included what pull requests
+members of the project would like reviewers to focus on during the
+upcoming week, when to disclose known DoS vulnerabilities for versions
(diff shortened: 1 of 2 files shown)


_posts/en/meetings/2018-06-21-meeting.md
```diff
@@ -0,0 +1,336 @@
...
+**Background:** the development branch ("master" branch) of Bitcoin Core
+includes code that allows users to dynamically load and unload
+individual wallets in multiwallet mode.  For example, you can have a
+"personal" wallet and a "business" wallet that can each be opened or closed
+separately.
+
+**Discussion ([log][log multiwallet persistence]):** Jonas Schnelli
+requested and introduced the topic, "I guess it's not ideal that loaded
+wallets need to be re-loaded after a Bitcoin Core restart---especially
+in pruning mode."
```
jnewbery: This could perhaps use more context. The risk here is that someone:

- starts bitcoind
- loads a wallet at runtime
- stops bitcoind
- restarts bitcoind

At this point, the previously dynamically loaded wallet will not be loaded, the tip will advance, but the wallet will not be notified of new transactions. Therefore when the wallet is used again, it would require a rescan to ensure that it hasn't missed txs. Jonas's comment about pruning is that if the node has pruned beyond the wallet's last processed block in the meantime, it's not possible to do a rescan.

_posts/en/meetings/2018-06-21-meeting.md
```diff
@@ -0,0 +1,336 @@
...
+[rwconf meeting 2018-06-07]: /en/meetings/2018/06/07/#command-line-argument-mapping
+
+**Conclusion:** "Okay, guess rw/config solves this, so /topic," said
+Schnelli.
+
+## Bech32x
+
+**Background:** several Bitcoin Core contributors have been working to
+create a new serialization format for backing up and recovering Bitcoin
+private keys, HD-wallet seeds, and HD-wallet extended private keys.  The
```
jnewbery: and extended public keys

PR #569, opened 2018-06-26
Bugfix: fix broken Markdown reference-style links

harding:
If you create a Markdown reference-style link whose target isn't defined, the link definition is rendered unchanged onto the resultant page.  E.g. [broken][broken link]

This PR adds a test that greps for those broken definitions and fixes all current instances of them.


harding:
Oh, I should've mentioned that it's pretty easy to review diffs in many languages if you review the same commit twice: 

1. Once using Google machine translation in Chrome, Chromium, or Firefox with Google Translate plugin for translation correctness, and 
2. Once in the untranslated mode for markup correctness, since the machine translation both adds and removes characters that would affect formatting.

This is part of the reason I separated the commit for the non-English content from the English content.  Here's a screenshot of part of that diff after machine translation; you can see that it's pretty easy to determine that the link reference text seems correct, although the formatting and even parts of the translation are changed from their original in ways that aren't representative of the actual diff:

_posts/zh_CN/releases/2016-10-27-release-0.13.1.md
```diff
@@ -99,7 +99,7 @@ Segwit 软分叉的激活是用 BIP9 的 versionbits 来实施的。Segwit 的 v
 日启动之后，节点们就会在 2016 个区块算力重新调整期的开始的时候追踪区块支持 Segwit 的信号。当这两个礼拜的重新调整期内，有
 95%的区块发出信号支持 Segwit，这个软分叉就会被锁定。然后过了另外 2016 个区块后，Segwit 就会被激活。 
 
-如果你想获得关于 Segwit 的更多信息，请参阅 [Segwit FAQ][]，和 Segwit [钱包开发者指导手 册][]  或是 [141][BIP141], [143][BIP143],
+如果你想获得关于 Segwit 的更多信息，请参阅 [Segwit FAQ][]，和 Segwit [钱包开发者指导手 册][segwit wallet developers guide]  或是 [141][BIP141], [143][BIP143],
```
jnewbery: I believe the text for the first link here should be 隔离见证的好处 rather than 'Segwit FAQ' and it should link to https://bitcoincore.org/zh_CN/2016/01/26/segwit-benefits/ rather than the English page.

harding: That wasn't a link changed in this PR, but I agree it would be an improvement.  Commit added to PR, thanks!

PR #571, opened 2018-07-01
Scripted diff: rename files in _includes with leading underscores

harding:
We have a number of files in the `_includes` directory whose file names start with a leading underscore, e.g. `_references.md`.  This leading underscore doesn't provide any feature, isn't a convention of Jekyll sites[1], is a little bit annoying to type, and is a lot annoying when editing with syntax highlighting where including a file that starts with an underscore causes the highlighter to think Markdown italics are in use for the remainder of the document.

I haven't bothered to rename the files to eliminate the leading underscore before because it seemed mostly to be just a personal peeve of mine.  However, I was recently working on a site by some BitcoinCore.org contributors and discovered that the habit has spread to that site, probably because they thought its use here was conventional.  So this PR cures patient zero with the hopes of preventing further infections.

I confirmed through diffing builds of the site before and after this commit that it doesn't introduce any content changes.  The rename script is in the commit comment; requires GNU coreutils.

---

[1] I suspect it's done as a misunderstanding of leading underscores in the top-level directory, where they do have meaning.  They have no meaning in subdirectories.


jnewbery:
utACK 1b790563902aee452e52bc1089af297cb8e027f0.

I've reviewed the script and verified that there are no `\b_\.html\b` and `\b_\.md\b` words in the `_include` directory. I haven't checked every line of diff.

jnewbery:
Approving this so it can be merged (since another PR builds on it).

@jonasschnelli @btcdrak - you may want to review this. It's a large delta, but it's scripted so should be an easy review.

PR #582, opened 2018-07-27
Releases: add 0.16.2

harding:
Adds release notes and a short blog post.  Previews:

- Blog: http://dg0.dtrt.org/en/2018/07/27/release-0.16.2/
- Release notes: http://dg0.dtrt.org/en/releases/0.16.2/
- Download page: http://dg0.dtrt.org/en/download/

Todo:

- [x] Update files to correct dates for release:
    - `_releases/0.16.2.md` - change date in YAML header
    - `cd _posts/en/posts/ ; rename 's/.*release/'$(date -u -I)'-release/' _posts/en/posts/2018-07-27-release-0.16.2.md ; git add .`
- [x] Add magnet URI for torrent to _releases/0.16.2.md
- [x] If release notes change, update the release file with the changes

Note: this PR is expected to fail Travis CI as it expects the binaries to be in /bin/, which they're not yet.  The CI should pass once if Travis is restarted after the binaries are in their final location.

PR #585, opened 2018-07-29
Bitcoin Core 0.16.2

laanwj:
can be merged if it passes travis

### DIFF
--- a/_releases/0.16.2.md
+++ b/_releases/0.16.2.md
@@ -1,0 +1,144 @@
+---
+title: Bitcoin Core 0.16.2
+id: en-release-0.16.2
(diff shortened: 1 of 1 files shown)


harding:
Oh, we already had #582 open.

harding:
Weird.  This fails because it says the torrent file is missing, but I see it listed in https://bitcoincore.org/bin/bitcoin-core-0.16.2/ .  But but, if I click on it, it 404s.

ken2812221:
Got 403 here
image

laanwj:
oh, sorry, will close and check the file
might be a permissions issue though my upload script is supposed to normalize those

PR #589, opened 2018-07-31
Add 0.16.2 RPC docs

harding:
Preview: http://dg0.dtrt.org/en/doc/0.16.2/

First commit is auto generated using the `contrib/doc-gen/generate.go` script and should be deterministic.  Second commit just updates a version number.

CC: @karel-3d


jnewbery:
Tested ACK 471fd33d281672bd74764d18373d4cdc3393a10e

I verified that the first commit is the result of running `generate.go`, and that the second commit correctly redirects to the 0.16.2 rpcs. I didn't manually check all of the text.

PR #592, opened 2018-08-06
Add Download link to japanese navigation menu

azuchi:
Add Download link to japanese navigation menu and modify and fix some navigation links to Japanese pages.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -285,12 +285,15 @@ ja:
           faq:
             title: "FAQ"
             url: "/en/about/sponsorship/faq/"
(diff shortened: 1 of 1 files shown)


harding:
80a6b23e0025f5b2676e943baba9fe2aaa36182e tested LGTM.  Thanks!

@azuchi while you're editing this file, did you want to provide translations for all the top-level menu items?  That appears to be what was done for other transactions, e.g. Simplified Chinese:

azuchi:
@harding thanks for the review! Following the advice, I translated the menu item to japanese.

PR #599, opened 2018-09-19
Verify commits: update allow-revsig list for expired key

harding:
@btcdrak's old key has expired, causing the verify-commits.sh script to fail and the website not to build normally.  Edit: this PR now updates the allow-revsig list as was indicated in https://github.com/bitcoin/bitcoin/pull/12822

~~This updates the trusted commit to point to the latest commit using that key so only subsequent commits are validated.  I believe the procedure is for people to check their own local repository and see if the new trusted commit id is a part of their copy of the master branch---even better if it's buried somewhere under a commit you PGP signed.~~

```text
user@devbco:~/bcco-master$ git show --show-signature 32b7869956e3a70025e32b69bab11fc262ef84d8
commit 32b7869956e3a70025e32b69bab11fc262ef84d8
gpg: Signature made Sat 16 Dec 2017 10:35:52 PM EST
gpg:                using RSA key C26B028A44F111A0
gpg: Good signature from "BtcDrak <btcdrak@gmail.com>" [unknown]
gpg: Note: This key has expired!
Primary key fingerprint: 912F D322 8387 123D C97E  0E57 D556 6241 A029 5FA9
     Subkey fingerprint: 8E75 26B0 B84F 2D55 8C5F  A39E C26B 028A 44F1 11A0
```

~~I'm just mimicking what was done [here](https://github.com/bitcoin/bitcoin/commit/7deba93bdc76616011a9f493cbc203d60084416f) and confirming that it resolves the problem with verify-commits.sh.  If there's a more appropriate solution, please let me know.~~

@MarcoFalke of the regular reviewers for this repository, you seem to have the best understanding of the github-merge.py/verify-commits.sh workflow, so your review would be especially appreciated.


MarcoFalke:
Separate from that it could make sense to run the check in travis (see https://github.com/bitcoin/bitcoin/blob/master/.travis/lint_06_script.sh#L23)

ken2812221:
Would it be better to follow bitcoin/bitcoin#12822?

MarcoFalke:
For me this seems like overkill. I am not objecting to fix it in that way, but it is mostly a waste of time to verify all commits down to the trusted root, when another commit (32b7869956e3a70025e32b69bab11fc262ef84d8, about one year old and signed off by all commiters after that day) is known to be verified, given that current master is verified. It is sufficient to verify commit 32b7869956e3a70025e32b69bab11fc262ef84d8 once and then never again. Everything else is just a waste of (compute) time.

Also note that a specially crafted merge commit can override all previous signatures except the one of the trusted root. So maybe bumping the trusted root is a good way to detect these, if they were not intentional.

harding:
1. Yeah, running the test in Travis is something I plan to PR as this expired key caused a recently-merged PR to not get automatically deployed to the website, a problem that we would've had a day to fix if we'd known about it earlier.

2. @ken2812221 thanks for referencing that PR, I hadn't seen that in my search for how upstream handles this situation.   Although both solutions seem fine to me, that seems like the official way we're supposed to handle revoked keys, so I'll revise this PR to use that solution.  (I agree with @MarcoFalke's logic, but I generally prefer to mirror upstream's policies in this repository for all cases where it applies reasonably well.)

harding:
As discussed above, this PR now puts the list of commits signed with BtcDrak's expired key into the allow-revsig-file so that their key-expiration warning is ignored (but those commits and signatures are otherwise checked).  The list was generated with the following command:

    git log --format="%G? %H" $(
      cat contrib/verify-commits/trusted-git-root
    )...HEAD 2>/dev/null | grep ^Y | sed 's/^Y //' \
    > contrib/verify-commits/allow-revsig-commits

This PR has an additional commit that updates the `gpg.sh` script to the lastest version from the upstream Bitcoin Core master branch to fix a bug with the allow-revsig-commits handing in the version in this repository.

With this change, I've verified that `verify-commits.sh`returns successfully.

nathsmaR:
Go

TheBlueMatt (approved):
Didn't check the allow-revsig-commits list but it doesn't matter much, the gpg.sh changes look like the ones from Bitcoin Core.

SatoshiNakamotogmx (approved):

PR #600, opened 2018-09-21
Posts: full disclosure for CVE-2018-17144

harding:
Only change from text I received is the excerpt and addition of bullet point markup.


sipa:
Looks good.

luke-jr:
This seems way too premature. Only 2% of the network has upgraded.

PR #603, opened 2018-09-21
CVE post: rephrase technical details

harding:
This is an alternative to #602 with the background placed a bit more in time-linear order and with some additional information about the apparent motivations for bitcoin#443 and bitcoin#9049.

CC: @TheBlueMatt

### DIFF
--- a/_posts/en/posts/2018-09-20-notice.md
+++ b/_posts/en/posts/2018-09-20-notice.md
@@ -29,7 +29,19 @@ However, it still remains critical that affected users upgrade and apply the lat
 Technical Details
 =================
 
-In Bitcoin Core 0.14, an optimization was added (Bitcoin Core PR #9049) which avoided a costly check during initial pre-relay block validation that multiple inputs within a single transaction did not spend the same input twice which was added in 2012 (PR #443). While the UTXO-updating logic has sufficient knowledge to check that such a condition is not violated in 0.14 it only did so in a sanity check assertion and not with full error handling (it did, however, fully handle this case twice in prior to 0.8).
+The original implementation of Bitcoin ensured that no transaction
+within a block spent an input that had already been spent on the same
+block chain.  However, it failed to ensure that the transactions it
+relayed and accepted into its memory pool only spent the same input
+once, so PR #443 was opened to add a rule that rejected transactions
+with duplicate inputs.  This became part of the 0.4 release of Bitcoin.
+
+This new rule was added to the function used to validate all
+transactions, whether they were unconfirmed transactions being relayed
+or transactions included in blocks, which meant that Bitcoin software
+was redundantly checking blocks for duplicate inputs twice.
+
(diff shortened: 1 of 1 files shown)


_posts/en/posts/2018-09-20-notice.md
```diff
@@ -29,7 +29,19 @@ However, it still remains critical that affected users upgrade and apply the lat
...
+within a block spent an input that had already been spent on the same
+block chain.  However, it failed to ensure that the transactions it
+relayed and accepted into its memory pool only spent the same input
+once, so PR #443 was opened to add a rule that rejected transactions
+with duplicate inputs.  This became part of the 0.4 release of Bitcoin.
+
+This new rule was added to the function used to validate all
+transactions, whether they were unconfirmed transactions being relayed
+or transactions included in blocks, which meant that Bitcoin software
+was redundantly checking blocks for duplicate inputs twice.
```
jnewbery: I think this could be made even clearer with something like:

```
redundantly checking blocks for duplicate inputs twice:
- once when validating the transaction
- once when updating the set of unspend outputs (UTXOs)
```
That would make the following section about 'UTXO-updating logic' clearer.

_posts/en/posts/2018-09-20-notice.md
```diff
@@ -29,7 +29,19 @@ However, it still remains critical that affected users upgrade and apply the lat
...
+
+In Bitcoin Core 0.14, an optimization was added (PR #9049) which eliminated this redundancy. While the UTXO-updating logic has sufficient knowledge to check that such a condition is not violated in 0.14 it only did so in a sanity check assertion and not with full error handling (it did, however, fully handle this case twice in versions prior to 0.8).
```
jnewbery: I think 'it did, however, fully handle this case twice in versions prior to 0.8' should either be expanded (which versions? How? Why was this error-handing changed?) or removed.

jnewbery:
ACK, this is clearer. A couple of suggestions inline.

harding:
@jnewbery added a commit that I believe addresses your concerns.  Note that I use the term "transaction database" instead of "UTXO set" because 0.4 was pre-ultraprune, so it actually stored all transactions back then.

jnewbery:
ACK 41a61a1528c55503974a319156e31c1a20191def. Looks great!

harding:
I'm not confident that this is entirely correct, specifically this sentence:

> However, [Bitcoin 0.1] failed to ensure that the transactions it
relayed and accepted into its memory pool only spent the same input
once, so PR 443 was opened to add a rule that rejected transactions
with duplicate inputs.

Because of that, I've been holding off merge until someone familiar with the motivations for bitcoin#443 could comment.  However, now that it's been three weeks since the publication of the notice, It think it's better to keep the original text of the CVE notice for historical preservation and so I'm closing this PR.

PR #610, opened 2018-09-26
Change announcements mailing list to RSS (at least temporarily)

harding:
I've been unable to contact @btcdrak, the administrator of the security and releases announcement mailing list, and it doesn't seem to be sending emails (certainly not reliably).  It'd be nice to set up a replacement, but right now other existing Bitcoin mailing lists are also in flux so there's no stable hosting available without creating a new dedicated solution.

This PR provides an alternative news source by recommending users follow an RSS feed for posts on the website tagged as announcements.  This isn't as good as email, but I think it's better than nothing (and certainly it's better than recommending people join a broken list).

Preview feed: http://dg0.dtrt.org/en/announcements.xml
Preview sign up page (menu->Contact->Announcements): http://dg0.dtrt.org/en/list/announcements/join/
Feed validation: https://validator.w3.org/feed/check.cgi?url=http%3A%2F%2Fdg0.dtrt.org%2Fen%2Fannouncements.xml

Pages that previously mentioned the email list have been updated to refer to the RSS feed; for the translations, that was done by deleting whole sentences with the help of Google Translate, so the text is a bit sparse (but seems clear to me).  The initial announcement blog post of the email list was translated into Chinese; I didn't see any reason to attempt retranslating an outdated blog post from two years ago, so I deleted it and redirected its URL to the corresponding English blog post (which is updated).

### DIFF
--- a/_includes/pages/list/announcement.html
+++ b/_includes/pages/list/announcement.html
@@ -1,6 +1,1 @@
-<form action="https://sendy.bitcoincore.org/subscribe" method="POST" accept-charset="utf-8" id="mailregform">
-    <input type="hidden" name="list" value="NkAKeRXqqhq763VNq4wM8921SA"/>
-
-    <input type="text" placeholder="Your email address" name="email" id="email"/>
-    <input type="submit" name="submit" id="submit"/>
-</form>
+[RSS](/en/announcements.xml)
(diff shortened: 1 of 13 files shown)


laanwj:
Concept ACK.
I was able to contact @btcdrak at some point and he told he was going to look into it, but AFAIK hasn't yet and it's still not working. We need an alternative solution, and this will do until we have some kind of e-mail notification up and running again.

PR #618, opened 2018-10-03
RPC Docs: add 0.17.0

karel-3d:
It's that time again :D  

Now the diff is a bit bigger (2k lines), with all the new RPC, so I won't paste it here

https://gist.github.com/karel-3d/e859f4b43f00344c60e948988aa21849


karel-3d:
The right menu looks really terrible for Wallet now :( since the number of RPC grows

But fixing that is a bigger redesign

PR #620, opened 2018-10-12
Add japanese translation for CVE-2018-17144

azuchi:


azuchi:
@harding oh, sorry. Gemfile.lock did not matter. I use `assume-unchanged`. thanks.

PR #623, opened 2018-10-25
trivial: Fixed typo 

merland:


karel-3d:
ACK

jnewbery:
I *think* the text is ok as is, since it's referring to multiple versioning systems, but I'll defer to the @harding stylebook on this one.

merland:
@jnewbery Thanks for weighing in. 
This clearly is an honest typing mistake. Nobody says "versions numbers", right?

harding:
It's wrong, probably an editing mistake on my part where I rearranged the sentence.  ACK.

PR #627, opened 2018-11-03
Add japanese translation for lifecycle

azuchi:


_posts/ja/pages/2016-01-15-lifecycle.md
```diff
@@ -0,0 +1,85 @@
...
+
+## メンテナンスリリース {#maintenance-releases}
+
+メジャーリリースのバグを修正するメンテンナンスの「マイナーリリース」を提供します。一般的なルールとして、メンテナンスリリースでは（コンセンサスルールを除いて）重要な新しい機能を導入していません。ただし、必要に応じてマイナーな機能を追加したり、ソフトフォークなどによるコンセンサスルールの変更をバックポートすることがあります。
+
+マイナーリリースには、0.11.1、0.11.2、0.12.1、0.12.2といった番号が付けられます。
+
+## コンセンサスルール {#consensus-rules}
+
+コンセンサスルールを変更する提案は、常に0.11.2、0.12.1などの最初のメンテナンスバージョンでリリースされます。これにより、メジャーリリースに比べてチェンジセットが少なくなるため、エンタープライズユーザーにとって提案の評価とテストが容易になります。またより慎重なアップデートパスをとるユーザーにとっては、コンセンサスルールの変更をより適切な時期に採用することができるようになります。
```
harding: Google translates this first sentence as,

> Proposals to change consensus rules will always be released in the first maintenance version, such as 0.11.2, 0.12.1.

The original text says,

> Proposals to change consensus rules are always shipped first in maintenance versions such as 0.11.2, 0.12.1 etc.

This is a subtle difference that could just be Google translating it weirdly, but I want to make sure it's translated correctly: the project doesn't guarantee that consensus-changing code will be in a x.y.1 release (the first maintenance version), just that it'll be in an x.y.something release (any maintenance version).  When I wrote that text, I was sure to select [0.11.2](https://bitcoincore.org/en/releases/0.11.2/) (BIP65/112/113 soft fork code) as one of the examples for the reason of making this clear.

PR #638, opened 2019-02-15
Remove sha256sum option not present in Debian

lucayepa:
The option --ignore-missing is not present in the Debian version of sha256sum. So "sha256sum --ignore-missing --check SHA256SUMS.asc" does not work.

### DIFF
--- a/_includes/templates/download.html
+++ b/_includes/templates/download.html
@@ -198,7 +198,7 @@
 
       <li><p>{{page.verify_download_checksum}}</p>
 
-        <pre class="highlight"><code>sha256sum --ignore-missing --check SHA256SUMS.asc</code></pre>
+        <pre class="highlight"><code>sha256sum --check SHA256SUMS.asc 2&gt;/dev/null | grep OK</code></pre>
 
         <p>{{page.checksum_warning_and_ok | replace, "$(SHASUMS_OK)", page.localized_checksum_ok}} <code>{{FILE_PREFIX}}-{{site.data.binaries.lin64}}: {{page.localized_checksum_ok}}</code></p></li>
 


harding:
@lucayepa what version of Debian are you using?  It works for me on current stable:

    $ cat /etc/issue
    Debian GNU/Linux 9 \n \l

    $ sha256sum --help | grep ignore-missing
          --ignore-missing  don't fail or report status for missing files

    $ dpkg -S $( which sha256sum )
    coreutils: /usr/bin/sha256sum

    $ dpkg -l coreutils | cat
    Desired=Unknown/Install/Remove/Purge/Hold
    | Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
    |/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
    ||/ Name           Version      Architecture Description
    +++-==============-============-============-=================================
    ii  coreutils      8.26-3       amd64        GNU core utilities

lucayepa:
Debian Jessie:
```
$ cat /etc/debian_version 
8.11
$ sha256sum --help | grep ignore-missing
$ dpkg -S $( which sha256sum )
coreutils: /usr/bin/sha256sum
$ dpkg -l coreutils | grep ii
ii  coreutils      8.23-4       amd64        GNU core utilities
```

harding:
@lucayepa Thanks for checking that.  Jessie is oldstable (initially released almost 4 years ago).  I'm a bit reluctant to change the current simple and clear command line to something more complex and abstruse to accommodate an older system that's probably primarily being used by advanced users anyway.  What do you think?

lucayepa:
I think you are right. I wonder if it is only related to Debian oldstable, or if there are other distributions still around. Usually Debian is slower to use new software, so maybe it is the only one. I've tried to find a clearer way, but I haven't found anything better than a pipe and a grep.

harding:
@lucayepa if more examples pop up of people unable to run the command because their sha256sum doesn't support the `--ignore-missing` option, I'll certainly support changing the instructions.  I'd probably do it like the macOS instructions we currently provide where we simply tell people to parse the output for themselves looking for the "OK".  That's ugly in its own way, but it's simple.

Thanks anyway for reporting this issue and including a proposed fix for it!

PR #640, opened 2019-02-15
Download page: remove link to laanwj's key; mention --refresh-keys

harding:
Recently, @laanwj's release signing key expired.  He has uploaded to key servers an updated key with a future expiration date, but the copy of the key hosted on this server is still the expired version.  Instead of updating it, this drops the link to it on the Download page and also adds a tiny note about `gpg --refresh-keys`  Preview (bottom text):

The verification instructions on the page all mention getting the key with `--recv-keys` and there are no other links to the key on this site, so I think this is an acceptable solution.  However, this PR leaves the actual expired key on the site in case any third party scripts are grabbing it automatically and just ignoring its expired status.

### DIFF
--- a/_includes/templates/download.html
+++ b/_includes/templates/download.html
@@ -80,7 +80,8 @@
     </p>
     <p class="downloadkeys">
       <span>{{ page.releasekeys }}</span>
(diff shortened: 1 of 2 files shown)


MarcoFalke:
ACK

PR #641, opened 2019-02-21
Remove mention of `contact@` and the slack

laanwj:
The `contact@` address for "For general enquiries and press" is unattended. No one is handling these since @btcdrak became inactive, so imo it's better to remove it.

As far as I know, no bitcoin core developers regularly attend the Slack, so remove it from contact information as well.

PR #643, opened 2019-03-04
Modify japanese about #641

azuchi:


harding:
utACK a2af2115da7efa9f8c70f5a521feeae2915c0435 thanks!

PR #648, opened 2019-04-11
annoucements: Mention bitcoin-core-dev mailing list

MarcoFalke:


MarcoFalke:
WIP

harding:
Adding the core-dev mailing list is a good idea, but I'd suggest the text mention that it's a discussion list (in theory, at least) not just an announcements list.  IMO, the problem with recommending users subscribe to discussion lists is that many of them aren't interested in the discussions, so they stop reading the posts and miss the actual announcements.  (Heck, not even several major Bitcoin protocol devs want to read the posts on the protocol discussion list.)

Maybe the new list infrastructure will allow us to create a third announcements-only list (e.g. every post must be manually approved) so we can restore the old list feature.  Alternatively, we could just use a commercial email service designed to deliver blasts (I think this has been opposed in the past based on concerns over sharing user emails with a commercial service; I personally think it's better than not having an announcements list at all); this can be combined with an [rss-to-email](https://mailchimp.com/features/rss-to-email/) feature if we wanted to automatically send readers posts using this website's announcement tag.

MarcoFalke:
I am not aware that it has been used for anything other than update reminders in the past, but if we want to have to separate mailing list I won't object.

Example
https://lists.linuxfoundation.org/pipermail/bitcoin-core-dev/2018-September/thread.html

harding:
Eh, https://lists.linuxfoundation.org/pipermail/bitcoin-core-dev/2016-April/000004.html  :smiley:

MarcoFalke:
Heh, good point.

laanwj:
In practice it's not used as a discussion list, to be honest I wouldn't really *like* it to be used for discussion. IRC and github together cover the project's communication quite well.

So I'm ok with officially turning it an announcement list where only specific people can post.

laanwj:
Sorry for waiting so long with this I only just realized I can do this myself :woman_facepalming: 

I've changed the settings of `bitcoin-core-dev` so that all messages are held for moderation. I've also updated the description that it's an announcement list.

_posts/en/pages/list/2016-03-10-announcements-join.md
```diff
@@ -9,7 +9,7 @@ version: 2
 ---
```
harding, line 9: Please increment the `version` field on line 8 so we can keep the translations in sync.

_posts/en/pages/list/2016-03-10-announcements-join.md
```diff
@@ -9,7 +9,7 @@ version: 2
...
 Receive notification of important security announcements and releases for Bitcoin Core.
 
-Subscribe to the RSS feed below.
+Subscribe to the RSS feed below or add your email address to the [bitcoin-core-dev][] mailing list.
```
harding, line 12: Need to add the following to the bottom of the file to use the global reference links:

    {% include references.md %}

harding (changes requested):
Two tiny changes needed.  Otherwise LGTM.  Thanks!

harding (approved):
Tested ACK fb9acf7e02bb02cd1fbbeb19edb6daa83a454d14 Thanks!

PR #649, opened 2019-04-11
Releases: add 0.18.0

harding:
Previews:

- http://dg0.dtrt.org/en/2019/05/01/release-0.18.0/
- http://dg0.dtrt.org/en/releases/0.18.0/
- http://dg0.dtrt.org/en/download/
- http://dg0.dtrt.org/en/doc/0.18.0/
- http://dg0.dtrt.org/en/announcements.xml

This adds the regular blog post and release notes for 0.18.0.  It departs from earlier release in a few small ways:

- RISC-V release added to download page; Win32 files removed 

- 0.18.0 added to the new announcements feed

- Updated RPC docs are included in this PR rather than a separate PR after release

- ~~Links to the 0.18.0 win32 files are explicitly ignored in the download linkchecker test~~  Edit: not necessary since it was [requested](https://github.com/bitcoin-core/bitcoincore.org/pull/649#issuecomment-482359033) to simply remove the links instead.

#### Todo

- [x] Update `_releases/0.18.0.md` with,
    - final release notes in the body of the document
    - actual release date in the YAML header
    - magnet URI in the YAML header
- [x] Rename `_posts/en/posts/2019-05-01-release-0.18.0.md` to include the actual release date in the file name

Note: In Travis, the `contrib/qa/test-binary-availability.sh` test is expected to fail until the final binaries have been uploaded. I suggest restarting the tests after upload and ensuring they pass before merging this; tests take about 5 minutes if Travis isn't busy.

Second note: if Win32 binaries are uploaded after release, revert the commit titled "Download page: remove 32-bit Windows links" to display them on this page.


achow101:
The download links for win32 on the downloads page should be removed as well.

luke-jr:
Maybe Windows becomes one column like macOS ("installer - zip" like "dmg - tar.gz")

harding:
@achow101 oh, is that what was agreed?  When @luke-jr said "neglect to post the binaries" I thought the goal was to have broken links.  I do like removing the links better (and it obviates the need for the test change), so I'll make that change.  I'll still do it in a separate commit so that it's easy to revert if it's decided to later upload the files.

harding:
Links to the win32 files have been removed.  The preview has been updated (links in OP).  As @luke-jr suggested, Windows has been become a single bullet with `exe` and `zip` links, both for win64.

harding:
This has been updated with the release notes at the time 0.18.0 was tagged as final and a release date assumption of May 1st.  If the release is on 5/1, the only change that should need to be made is adding the magnet URI to the release file (see OP).

laanwj:
Thanks !
Added commit with magnet link.

PR #650, opened 2019-05-02
releases: Change windows default to win64

laanwj:
Set win64 as default for windows users.
It has better performance, supports higher dbcache values, and so on.
Surprised this wasn't the case already.


gmaxwell:
utACK

luke-jr:
utACK

fanquake:
utACK 6a2934f

PR #639, opened 2019-02-15
Remove download references to PPA

MarcoFalke:
Closes #635

PR #653, opened 2019-05-20
download: Add link to snap store

MarcoFalke:


MarcoFalke:
We use ubuntu either way to build the binaries, so this shouldn't make anything worse for non-ubuntu linux distros that have the snap package manager.

harding:
Tested, no problems with the code.  Thanks!

When when #639 removed the old PPA, it also removed the [notice about non-reproducibility](https://github.com/bitcoin-core/bitcoincore.org/pull/639/files#diff-6f7e63e1575174cab9d999ed5c59233dL104).  My understanding is that the snap includes the gitian-verified binaries; is it possible for the user to verify that given the checksum file also linked from the download page?  If so, I'd like to see either this PR include verification instructions (similar to those already on the page for other binaries) or an issue opened in this repo so that we don't forget that those instructions should be written after this is merged.  If verification is not possible (or impractical), then I'd like to see a warning about non-reproducibility similar to the old warning.

luke-jr:
Theoretically, it should be possible to generate the actual snap packages in gitian, I think? (Although when I tried to do so, I failed...)

harding:
Tested ACK fafa3bdd95399115a276e826f6c372ccdc657861 Thanks!

Issue #607, opened 2018-09-24
Is the sponsorship programme still alive?


karel-3d:
I have randomly stumbled on this sub-page on bitcoin core website.

* https://bitcoincore.org/en/about/sponsorship/programme/
* https://bitcoincore.org/en/about/sponsorship/faq/
* https://bitcoincore.org/en/2016/04/04/announcing_sponsorship_programme/

These webpages have been around since 2016, but I have not seen any wide coverage of them and I haven't heard about anyone participating, or any news.

Is the programme still alive?

harding:
I was hoping someone else with more information would comment on this issue.  I don't have any information and I haven't heard of any activity being funded through the program.  If the program is actively funding any efforts, we should probably do a better job of publicizing that.

jnewbery:
I don't know anything about these programs. Perhaps @btcdrak does?

These pages were added without a PR or discussion, so it's difficult to excavate the history around them. It seems to me that the programme was launched with good intentions, but fundraising and matching sponsors to developers seems like a hard thing to maintain, and probably outside the scope of what Bitcoin Core should be doing. Without the historic record of why these pages were added, it's difficult to know the rationale.

@laanwj, @jonasschnelli - anything to add here?

luke-jr:
I haven't directly seen any funding from this programme, but perhaps it is a source of funds used for coredev.tech meetings? I'm not really sure.

MarcoFalke:
I don't know if this program is still alive, but lately many companies have directly hired (and thus sponsored) developers to work full-time on Bitcoin Core or related technologies.

moneyball:
Should this page be removed if it isn't accurate or active? Alternatively, what do people think about linking to organizations that hire or sponsor Core development, or are actively seeking to do so?

MarcoFalke:
@moneyball Sounds good to me!

harding:
+1 on removing the page.  I'm meh on linking to organizations that sponsor Core dev for the same reason that we previously removed the page about Core developers.

jnewbery:
Concept ACK removing the sponsorship pages

MarcoFalke:
Would Bitcoin Ops be a more appropriate place for the page? They appear like the glue between protocol development and organizations.

jnewbery:
> Would Bitcoin Ops be a more appropriate place for the page? They appear like the glue between protocol development and organizations.

[...]

PR #657, opened 2019-05-30
sponsorship: Remove inactive program

MarcoFalke:
There is no indication that this program was ever active, so remove it to avoid spreading wrong information.


MarcoFalke:
See also discussion in https://github.com/bitcoin-core/bitcoincore.org/issues/607#issuecomment-424767201

@laanwj @jonasschnelli

jnewbery:
ACK fa05d55191e8e85356872b61f5bbe481738ecb0a

> With determination, no mountain is insurmountable.

_- MarcoFalke, having finally removed the sponsorship program_

laanwj:
ACK https://github.com/bitcoin-core/bitcoincore.org/pull/657/commits/fa05d55191e8e85356872b61f5bbe481738ecb0a

I don't think this was ever used, and might give people false expectations. Please remove it.

PR #659, opened 2019-06-07
Update EOL dates for 0.16 and 0.17

MarcoFalke:
EOL dates for 0.16 and 0.17 have not been announced, so do that by bumping the date of the previous version by 6 months.

Also, remove historic release dates for software version 0.8 -- 0.14, since we do not list the release dates for versions 0.7 and before.


harding:
This seems fine to me.  Do these dates need to be discussed somewhere?

jnewbery:
Concept ACK updating the dates.

Do we have a policy for how long the period should be between Maintenance End and EOL?

MarcoFalke:
Not really. And even 0.17. isn't maintained anymore

jnewbery:
Dates seem reasonable to me, but perhaps we should gather ACKs in the next IRC meeting?

laanwj:
ACK, looks good to me

MarcoFalke:
Looks like there haven't been any objections

jnewbery:
ACK cb223fca9f5dd1b69bb14cbaadd983ea12655370

PR #662, opened 2019-07-03
Add japanese translation for about and dev pages

azuchi:


harding:
Tested ACK 8f98a0a80786b072998faa921aff7cd5f9eaa999 Tested by previewing locally and reviewing using Google Translate

Issue #670, opened 2019-09-03
Delete unused transifex project


MarcoFalke:
There is a project on transifex, which is unused according to https://github.com/bitcoin/bitcoin/issues/16754#issuecomment-526373630. Thus, it should be deleted or deactivated to not waste translators time.

https://www.transifex.com/bitcoincore/bitcoincoreorg/dashboard/

harding:
Yes, please delete/deactivate.  I don't have access to delete the project on transifex, so I'm assigning to @laanwj

laanwj:
Done

PR #671, opened 2019-09-03
doc: Remove link to unused transifex

MarcoFalke:
See also #670


harding:
ACK fa19e77933a0df2a64bd890edc383960264a821d

I think it's ok to merge this now with #670 still open, but I'll wait a couple days in case #670 gets any discussion.

PR #685, opened 2019-11-25
Add torrent link to release page

Sjors:
We currently only show the torrent link on the download page, which only covers the most recent version.

This adds a torrent link on each release page:

Schermafbeelding 2019-11-25 om 17 04 46

I applied this to 0.19.0.1, 0.18.1 and 0.17.1; probably not worth doing it for older versions.

### DIFF
--- a/_includes/download_release.html
+++ b/_includes/download_release.html
@@ -1,0 +1,8 @@
+{% capture RELEASE %}{% for subver in page.release %}{{subver}}{% unless forloop.last %}.{% endunless %}{% endfor %}{% endcapture %}
+{% capture PATH_PREFIX %}/bin/bitcoin-core-{{RELEASE}}{% endcapture %}
+{% capture FILE_PREFIX %}bitcoin-{{RELEASE}}{% endcapture %}
+{% assign magnet=page.optional_magnetlink %}
+<https://bitcoincore.org{{ PATH_PREFIX }}/>
+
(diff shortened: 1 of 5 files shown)


_includes/download_release.html
```diff
@@ -0,0 +1,8 @@
+{% capture CURRENT_RELEASE %}{% for subver in page.release %}{{subver}}{% unless forloop.last %}.{% endunless %}{% endfor %}{% endcapture %}
+{% capture PATH_PREFIX %}/bin/bitcoin-core-{{CURRENT_RELEASE}}{% endcapture %}
+{% capture FILE_PREFIX %}bitcoin-{{CURRENT_RELEASE}}{% endcapture %}
+{% assign magnet=page.optional_magnetlink %}
+<https://bitcoincore.org{{ PATH_PREFIX }}/>
```
harding: I haven't tested this locally, but I think this code will give you the `.torrent` for the most recent release (e.g. 0.19.0.1 even if you're looking at the 0.17.1 page).  It should give you the correct magnet link.

Sjors: It doesn't; it uses `page.release`. But I renamed the confusingly named `CURRENT_RELEASE` to `RELEASE`.

laanwj:
Concept ACK

harding:
Tested ACK ddf22dcd0010c640f7175009a8dffdedae302ac1

PR #686, opened 2019-11-25
Bump Ruby, Jekyll and jQuery

Sjors:
Builds on top of #684
* Ruby 2.2.2 is end of life, so bumping to the latest and greatest 2.6.5
* `bundle update` mainly:
  * bumps Jekyll from 3.0.2 to 3.8.5, which required small tweaks to two files, but otherwise seems to behave
  * bumps html-proofer from 2.6.4 to 3.14.1 which is a bit more strict, hence the link fixes
* bumping jQuery from 1.9.1 to 3.4.1, from here: https://jquery.com/download/ (compressed, production jQuery 3.4.1), I didn't try the slim build


Sjors:
One benefit of updating stuff is that we're more likely to get automated pull requests from Dependabot for security vulnerabilities https://dependabot.com/blog/hello-github/

harding:
Any changes to the build deps need to be tested and deployed on the build server, which has historically been the blocker for this type of thing.  E.g. #480 

We already get spam from dependabot, e.g. #675 and #678.  Its nagging is almost entirely irrelevant to a statically built site.

harding:
(Sorting RPC in descending order is a nice improvement though, so I wouldn't mind that as an independent change.)

Sjors:
I split RPC doc sequence into it's own PR: #689, which can go either before or after this.

I also split the broken links fixes into #688, which should be merged before this (and is still included, so Travis passes).

TheBlueMatt:
The machine is just some stock-ish Debian Box. As long as it can apt upgrade to whatever is required that’s fine.

laanwj:
> We already get spam from dependabot, e.g. #675 and #678. Its nagging is almost entirely irrelevant to a statically built site.

Yes, I've never really understood what 'security vulnerabilities' mean in this context.

In the case of javascript libraries such as jquery, I guess there could be client-side vulnerabilities. But that seems mostly important when there is communication with third-party sites.

Sjors:
On the debian box, you may want to install a Ruby version manager. [Rbenv](https://github.com/rbenv/rbenv#installation) is low-touch and doesn't need sudo. But if you already use [RVM](https://rvm.io) just stick with that.

Make sure 2.2.2 is present: `rbenv install 2.2.2` and then test the current setup, or maybe #684

Then do `rbenv install 2.6.5` followed by `gem install bundler` and `bundle install` and check if this PR still works (I would assume it does, because Travis passes).

Sjors:
It should be noted that installing old Ruby versions will become increasingly tedious for developers: https://github.com/rbenv/ruby-build/issues/377#issuecomment-549717014

TheBlueMatt:
Note that box currently runs ruby 2.3.3 (and carries a patch to this repo to use 2.3.3). We'll have to upgrade debian it looks like to get new ruby? Or maybe we can just install some ruby2.6 package?

Sjors:
Not sure which versions are available for Debian.

The cool kids use Rbenv or RVM to get the right version: https://github.com/bitcoin-core/bitcoincore.org/pull/686#issuecomment-559469307

You could also compile from source: https://www.ruby-lang.org/en/documentation/installation/#installers

laanwj:
What's wrong with using Ruby 2.3.3?

Sjors:
It's also end of life: https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/

There's nothing specific in 2.6 that we need, but the longer we can delay having to do this again the better.

harding:
The default version of ruby in Debian stable is 2.5.1.  That seems recent enough?

```
$ dpkg -l ruby
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-========================================================================
ii  ruby           1:2.5.1      amd64        Interpreter of object-oriented scripting language Ruby (default version)
```

Sjors:
Ruby 2.5 is fine by me. Not sure if 2.5.1 is such a good idea; 2.5.7 is the most recent, with all sorts of fun CVE fixes: https://www.ruby-lang.org/en/news/2019/10/01/ruby-2-5-7-released/

Relying on the system version of Ruby doesn't seem like a good idea.

As @laanwj wrote:
> Yes, I've never really understood what 'security vulnerabilities' mean in this context.

This is true, although the sites content could be changed in someone sneaks an RCE into the build pipeline via some tainted markdown, that seems rather non-trivial.

TheBlueMatt:
Debian-Ruby-2.5.1 is almost certainly different from Ruby-2.5.1, especially after they release security notices. I dont see the issue in sticking with distro ruby, unless there's a strong reason why not to, it means updates are easier to handle (ie in one place) and integrates more cleanly with the rest of the system (which makes maintenance, especially part-time or with multiple people poking at it) much, much easier. Someone (or I) can update that VM to buster when this gets merged.

Sjors:
Switched to 2.5.1

jnewbery:
Concept and lightly tested ACK. Ruby 2.2.2 is ancient and is getting harder to install/use with recent distributions (see https://github.com/rvm/rvm/issues/4371 for example).

@TheBlueMatt - what needs to happen on the servers to allow us to use 2.5.1? Anything I can do to help?

TheBlueMatt:
Note that the repo is built on 2.3.3, not 2.2.2 (and carries a patch to build on 2.3.3), as its the default on debian oldstable (apparently?). I'm happy to switch it over to whatever is preferred when it makes sense, but obviously will require coordination in terms of merge-time.

jnewbery:
Thanks Matt. Are you ready for this to be merged?

harding:
I diffed the generated site output with and without this patch and there's a lot of changes.  Most of these don't matter (e.g. minor changes to the tags used by the syntax highlighter and the opengraph HTML tags), but some of the changes are unfortunate such as changing subhead anchor links (e.g. a header of `## Foo_Bar_Baz` used to get an id of `foobarbaz` but now gets `foo_bar_baz`).

It also seems to reduce the amount of information in the sitemap.xml file, e.g.

```diff
       <url>
         <loc>https://bitcoincore.org/en/doc/0.20.0/rpc/zmq/getzmqnotifications/</loc>
         
-        <lastmod>2020-08-12T08:02:26+00:00</lastmod>
+        <lastmod></lastmod>
         
         
       </url>
```

I don't see anyone bothering to fix any of those issues, so I guess we should just bite the bullet.  From what I see, the worst case is some incoming links might not point to the correct anchor (but people should still end up on the correct page).  So...

Tested ACK c6b281789683734c1cc4cbbbeed900215787eaf9  (reviewed code, skimmed very large diff of site build with this merged on master versus just master, verified hash of updated jquery matches [upstream](https://code.jquery.com/jquery-3.4.1.min.js))

TheBlueMatt:
Yea, let me coordinate with @harding so we can do it live.

PR #695, opened 2020-02-04
download: Remove i686 Linux link

MarcoFalke:
There is only one known user of the i686 linux binary, so it is planned to be removed from the build and quality assurance pipeline. The last user (or users) is encouraged to compile locally with `make check && ./test/functional/test_runner.py` to run the tests on their target system. Using clang, this is possible even on systems with 1 GB of ram. See https://github.com/bitcoin/bitcoin/issues/17504

Prepare this website and remove the download link from it early.


harding:
What's the goal for merging this, before 0.19.1 or before 0.20.0?

MarcoFalke:
I'd say before 0.19.1, so that everyone has a final chance to complain before the binaries are no longer produced and uploaded.

laanwj:
0.20.0 earliest

laanwj:
After discussion on IRC I'm fine with doing this as soon as possible. It serves as an advance warning, and given that no one really seems to care that these are going away we might as well stop building them for 0.20.

_includes/templates/download.html
```diff
@@ -251,21 +249,16 @@ <h2 style="text-align: center">{{page.build_reproduction}}</h2>
 <script type="text/javascript">
 var os = 'windows64';
 if (navigator.userAgent.indexOf('Mac') != -1) var os = 'mac'
-else if (navigator.userAgent.indexOf('Linux') != -1) {
-	var os='linux32';
```
laanwj: I guess we don't need this check for ARM, as the detection assumes Linux is always x86?

MarcoFalke: I think we assume that only x86 can run a browser :grimacing:

harding: Created an issue: https://github.com/bitcoin-core/bitcoincore.org/issues/696

harding:
Tested ACK fa73ac156faba562a023b2be474217c2c23b6f4e

harding:
From the weekly meeting:

```text
<wumpus> yes, so let's remove it from the website, but still build x86_32 binaries for 0.19.1
<MarcoFalke> +1
<luke-jr> sgtm
[...]
<emilengler> ack
```

Will merge momentarily.

PR #697, opened 2020-02-25
update estimated data directory size to 300 Gb

RandyMcMillan:
The current actual estimate is 280 Gb.
Rounding up per instructions.


harding:
ACK 21cdcab18ae93dcc827f672c99ce46d38c4b71ee Thanks!

PR #706, opened 2020-06-07
Add Onion-Location meta tag

RiccardoMasutti:
Starting with the release of [Tor Browser 9.5](https://blog.torproject.org/new-release-tor-browser-95), websites can have their alternate .onion addresses advertised to Tor desktop users who have the 'Onion Location' option enabled.

Sites that add the .onion address advertisement HTTP header can prompt their visitors to switch to a version delivered using the Onion service for improved security.

Final result:
ezgif-2-354636e604ed

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -11,6 +11,7 @@
 <title>{{ site.title | xml_escape }}{{ custom | xml_escape }}{% if title %} :: {{ title | xml_escape }}{% endif %}</title>
 {% if page.excerpt %}<meta name="description" content="{{ title | strip_html | xml_escape }}">{% endif %}
 <meta name="keywords" content="{{ page.tags | join: ', ' }}">
+<meta http-equiv="onion-location" content="http://6hasakffvppilxgehrswmffqurlcjjjhd76jgvaqmsg6ul25s7t3rzyd.onion{{ page.url }}" />
 {% if page.author %}
   {% assign author = site.data.authors[page.author] %}{% else %}{% assign author = site.owner %}
 {% endif %}


TheBlueMatt:
Hmm, I don't see any references to it working with a meta tag on that blog post, are we sure that works and not just an HTTP header? (We can add an HTTP header, too, but more manual stuff on the server config is much less ideal than a meta tag, I think).

_includes/head.html
```diff
@@ -11,6 +11,7 @@
 <title>{{ site.title | xml_escape }}{{ custom | xml_escape }}{% if title %} :: {{ title | xml_escape }}{% endif %}</title>
 {% if page.excerpt %}<meta name="description" content="{{ title | strip_html | xml_escape }}">{% endif %}
 <meta name="keywords" content="{{ page.tags | join: ', ' }}">
+<meta http-equiv="onion-location" content="http://6hasakffvppilxgehrswmffqurlcjjjhd76jgvaqmsg6ul25s7t3rzyd.onion" />
```
TheBlueMatt, line 14: Also, does this result in redirections from other pages to / ?

RiccardoMasutti: No, it redirects to the .onion homepage

TheBlueMatt: Wait, hmm? I mean if you go to bitcoincore.org/downloads, will it then redirect you to bitcoincore.onion/ instead of bitcoincore.onion/downloads?

RiccardoMasutti:
@da2x @TheBlueMatt maybe `<meta http-equiv="onion-location" content="http://6hasakffvppilxgehrswmffqurlcjjjhd76jgvaqmsg6ul25s7t3rzyd.onion{{ page.url }}" />` would be a cleaner solution. Don't you think?

RiccardoMasutti:
@TheBlueMatt @harding can you take a look?

harding:
This doesn't seem to work for me.  I uploaded a preview of the site (since normal localhost previews don't work with Tor) to http://dg0.dtrt.org/ and visited using the latest Tor browser (9.5.4 for GNU/Linux).  I can confirm the tag is in the HTML source, but I don't see any button or message or anything.

da2x:
@harding it only works on HTTPS.

RiccardoMasutti:
> @harding it only works on HTTPS.

Exacly.

> For the header to be valid the following conditions need to be fulfilled:
> 
> - The Onion-Location value must be a valid URL with http: or https: protocol and a .onion hostname.
> - The webpage defining the Onion-Location header must be served over HTTPS.
> - The webpage defining the Onion-Location header must not be an onion site.

harding:
@da2x @RiccardoMasutti that makes sense.  I installed a certificate, retested, and it works.

ACK d556c10ab53c8cb5e2704b930ef31b4fa2d46344 Reviewed code (all 1 line), checked site diff, and tested with Tor Browser via a temporary deployment.

Sjors:
Post merge tACK

PR #707, opened 2020-06-23
Segwit benefits: Retire "Signing of input values" section

luke-jr:
Since this feature is no longer viable in light of CVE-2020-14199, remove it from the page (as we did previously for compact fraud proofs).

### DIFF
--- a/_posts/en/posts/2016-01-26-segwit-benefits.md
+++ b/_posts/en/posts/2016-01-26-segwit-benefits.md
@@ -71,22 +71,6 @@ The modified hash only applies to signature operations initiated from witness da
  * [Proposal to limit transactions to 100kB](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2015-July/009494.html)
  * [Bitcoin Classic commit on 0.11.2 branch adding additional consensus limit on sighash bytes](https://github.com/bitcoinclassic/bitcoinclassic/commit/842dc24b23ad9551c67672660c4cba882c4c840a)
 
-## Signing of input values
-
-When a hardware wallet signs a transaction, it can easily verify the total amount being spent, but can only safely determine the fee by having a full copy of all the input transactions being spent, and must hash each of those to ensure it is not being fed false data. Since individual transactions can be up to 1MB in size, this is not necessarily a cheap operation, even if the transaction being signed is itself quite small.
-
-Segwit resolves this by explicitly hashing the input value. This means that a hardware wallet can simply be given the transaction hash, index, and value (and told what public key was used), and can safely sign the spending transaction, no matter how large or complicated the transaction being spent was.
-
-### Who benefits?
-
-Manufacturers and users of hardware wallets are the obvious beneficiaries; however this likely also makes it much easier to safely use Bitcoin in small embedded devices for "Internet of things" applications. 
-
-This benefit is only available when spending transactions sent to segwit enabled addresses (or segwit-via-P2SH addresses).
-
-### Further information
-
- * [BIP 143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki)
-
 ## Increased security for multisig via pay-to-script-hash (P2SH)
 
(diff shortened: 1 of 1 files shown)


_posts/en/posts/2016-01-26-segwit-benefits.md
```diff
@@ -71,22 +71,6 @@ The modified hash only applies to signature operations initiated from witness da
  * [Proposal to limit transactions to 100kB](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2015-July/009494.html)
  * [Bitcoin Classic commit on 0.11.2 branch adding additional consensus limit on sighash bytes](https://github.com/bitcoinclassic/bitcoinclassic/commit/842dc24b23ad9551c67672660c4cba882c4c840a)
 
-## Signing of input values
-
-When a hardware wallet signs a transaction, it can easily verify the total amount being spent, but can only safely determine the fee by having a full copy of all the input transactions being spent, and must hash each of those to ensure it is not being fed false data. Since individual transactions can be up to 1MB in size, this is not necessarily a cheap operation, even if the transaction being signed is itself quite small.
```
Sjors, line 76: Or just add with emphasis: _but only if it can sign all inputs_ and refer to the CVE

Sjors:
As long as you sign all inputs, a reasonable scenario for single-sig wallets, `CVE-2020-14199` isn't a problem. I would prefer to explain that, rather than drop the whole section.

luke-jr:
@Sjors My understanding is that this affects even if you sign all the inputs. Am I wrong? Why? BIP143 only ever signs the input amount for its own input.

Sjors:
The way I understand [the issue](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-August/014843.html) is as follows: SegWit commits to the input amounts _for the inputs you sign_. The attack consists of having at least two inputs, fooling the hardware wallet to think the other input is not theirs. That allows the attacker to fake the amount for the other input.

Adding a UTXO of the previous transaction reveals the amount for the inputs you're not signing. But if you sign all inputs, then the transaction would also be valid. Though come to think of it: this requires the device knowing that the input is valid. Otherwise, it would just produce an invalid signature for the other input, which the attacker can discard. In that case, we should indeed drop the section.

instagibbs:
ACK, I think the protections it gives is very subtle and should generally not be relied upon

harding:
@Sjors there's two different (but related) attacks in the email you linked.  In [one attack](https://bitcoinops.org/en/newsletters/2020/05/13/#request-for-an-additional-taproot-signature-commitment), the attacker lies about which inputs the wallet owns so that the wallet doesn't realize that it's added more input value that it receives output value.  In the [other attack](https://bitcoinops.org/en/newsletters/2020/06/10/#fee-overpayment-attack-on-multi-input-segwit-transactions), which is the basis for this removal PR, the attacker mixes truth and lies about the amounts so that the implicit change amount is much higher than the wallet thought it was.  Both attacks require signing the same transaction at least twice.

harding:
Untested ACK.  I didn't build a preview site but the diff LGTM and the change seems acceptable.

harding:
Tested ACK e92c6a44e770ac7ba115a08f331ee9c4c0aad7c0

gmaxwell:
Meh. I don't oppose the removal but it's still a benefit: the attack required against the segwit behaviour is much harder and requires an almost (but not quite) contrived user behaviour of the user voluntarily signing multiple distinct transactions making a payment redundantly.  If the user can be tricked into that, they can also be tricked into paying the destination multiple times.

As a general principle the protocol should be as de-footgunned as possible so the bip143 behaviour clearly isn't enough, but it's still an advance over the pre-bip143 behaviour.

luke-jr:
@gmaxwell But the user may see two 0.00001 BTC sends, and figure "meh, it's only 0.00002 BTC even if they're gaming me" when in reality the entire wallet is being emptied...?

gmaxwell:
Being sent to fees, which is it's own pita to deal with.  But yes, if not for stuff like that it would just be a non-issue.  It isn't a non-issue, but it's still better than the original behaviour.

PR #708, opened 2020-07-07
Update segwit benefit japanese page to reflect #707

azuchi:


harding:
Tested ACK 9b1725e525c85fbf0442a1f55462a26bab7fb764 (using Google Translate).  Thanks!

PR #719, opened 2020-08-19
About page: link to all contributors, not just recent

harding:
Closes #718 

This creates a new link definition, rather than changing the old one, because the old one is used in two translations.

PR #721, opened 2020-09-11
Updated japanese translation contributor list on about page

azuchi:
Updated japanese translation contributor list on about page to follow #719


harding (approved):
Tested ACK 1facb111b1aa29a214860672c09438af31a5cc51 using Google Translate.  Thanks!

PR #723, opened 2020-09-29
Create Tor Onion Location blog post

RiccardoMasutti:
Announce #706


harding:
I don't think this is necessary.  If people use Tor browser, they'll be notified.  If not, we already have a post announcing the existence of the hidden service, https://bitcoincore.org/en/2020/03/27/hidden-service/

laanwj:
I also don't really see why adding a HTTP header would need an announcement
post. End users have to do exactly nothing with it.

PR #726, opened 2020-10-20
Fix: Wrong file naming

awesome-doge:
Old Simplified Chinese translation in the past, with naming errors

### DIFF
--- a/_posts/zh_CN/2017-03-13-performance-optimizations-1.md
+++ b/_posts/zh_CN/2017-03-13-performance-optimizations-1.md
@@ -2,8 +2,8 @@
 layout: post
 type: posts
 lang: zh_CN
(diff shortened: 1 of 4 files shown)


harding:
Why do you think these files have the wrong names?

harding:
I agree that the language part doesn't need to be in the filename or the id tag, but is this causing any problems?  If it's not causing a problem, it usually doesn't need to be changed for old content.

awesome-doge:
If you think the extra language tags are not important, I will restore that commit.

MarcoFalke:
Is this still relevant?

awesome-doge:
I think it's better to close it first, I will handle all the problems inside in another pr
https://github.com/bitcoin-core/bitcoincore.org/pull/727

PR #729, opened 2020-11-26
Bump Approx disk size to 400 Gb

RandyMcMillan:

### DIFF
--- a/_data/stats.yml
+++ b/_data/stats.yml
@@ -1,7 +1,7 @@
 ## Size of the Bitcoin Core data directory running with default settings.
 ## Suggest rounding up by about 10% so it doesn't need to be updated
 ## too often.
-datadir_gb: "300"
+datadir_gb: "400"
 
 ## Size of the datadir with the maximum allowed pruning enabled.
 ## Suggest rounding up a little bit to give room for growth in the UTXO


Sjors:
400 should be enough for now: https://github.com/bitcoin/bitcoin/blame/v0.21.0rc2/src/chainparams.cpp#L108

RandyMcMillan:
@Sjors done

Sjors:
ACK 2868084

jonasschnelli:
My current blocks dir is 332GB. I think 400gb is enough for now but that number should bump soon again.
It's a bad experience if people buy disk space according to this number then figure "weeks" later out that its not enough to sync further (without prune).

harding:
@jonasschnelli I think it's fine.  The variable is used with this text on the [Download page](https://bitcoincore.org/en/download/):

> Bitcoin Core requires a one-time download of about 300GB of data plus a further 5-10GB per month. By default, you will need to store all of that data, but if you enable pruning, you can store as little as 6GB total without sacrificing any security. For more information about setting up Bitcoin Core, please read the full node guide.

harding:
ACK 2868084e467bb08fd644dd43586ebb5485516f41

PR #730, opened 2020-11-28
Update scaling term

murtyjones:
This PR updates `linear` to `quadratic` based on my impression that the sighash scaling problem SegWit solves is on the order of O(n<sup>2</sup>) and that Segwit reduces the complexity to O(n).

### DIFF
--- a/_posts/en/posts/2016-10-28-segwit-costs.md
+++ b/_posts/en/posts/2016-10-28-segwit-costs.md
@@ -148,7 +148,7 @@ In support of existing efforts to pay down technical debt by refactoring and imp
 
 ### Mitigation
 
-Bitcoin already suffers from some significant design debt, and segwit is specifically designed to reduce the impact of some of this debt (notably transaction malleability, linear scaling of signature hashing, and signing of input values).
+Bitcoin already suffers from some significant design debt, and segwit is specifically designed to reduce the impact of some of this debt (notably transaction malleability, quadratic scaling of signature hashing, and non-signing of input values).
 
 The script versioning method provided by segwit provides an elegant way of allowing future soft-fork updates to further reduce design debt, including by fixing bugs in existing opcodes (such as CHECKMULTISIG), re-enabling disabled opcodes (such as CAT), or switching to superior verification methods (such as Schnorr signatures, or aggregate signatures).
 


gmaxwell:
You might want to also make the "signing of input values" part phrased in terms of the flaw. e.g. non-signing of input values.

harding:
@murtyjones I see you've :+1:'d @gmaxwell's suggestion.  Did you want to update this PR with that change?  (I think, without that additional edit, the parenthetical would be just as confusing as it is now.)

CC original text author @ajtowns

murtyjones:
> @murtyjones I see you've 👍'd @gmaxwell's suggestion. Did you want to update this PR with that change? (I think, without that additional edit, the parenthetical would be just as confusing as it is now.)
> 
> CC original text author @ajtowns

Just added it. Also made the same change one other place in the text where I noticed the same issues

_posts/en/posts/2016-10-28-segwit-costs.md
```diff
@@ -298,8 +298,8 @@ Either approach to a hard-fork would make it possible to simultaneously drastica
 
 Many of the benefits of segwit could logically be separated into independent changes, and evaluated and deployed separately. The implementation requirements for the various features are, however, closely related:
 
- * Linear scaling of sighash operations: the CHECKSIG and CHECKMULTISIG opcodes need replacement.
- * Signing of input values: the CHECKSIG and CHECKMULTISIG opcodes need replacement.
+ * Quadratic scaling of sighash operations: the CHECKSIG and CHECKMULTISIG opcodes need replacement.
+ * Non-signing of input values: the CHECKSIG and CHECKMULTISIG opcodes need replacement.
```
harding, line 302: I don't think these should be changed.  The text before the colon describes the segwit feature being implemented (linear sig hashing, signing input values, 128-bit collision resistance for multiparty multisig, and more script upgrade paths); the text after the colon describes the change that needs to be made to the protocol.

murtyjones, line 302:
```suggestion
 * Linear scaling of sighash operations: the CHECKSIG and CHECKMULTISIG opcodes need replacement.
 * Signing of input values: the CHECKSIG and CHECKMULTISIG opcodes need replacement.
```

murtyjones: Good point, updated it

ajtowns:
Makes sense. Maybe squash the commits though :)

harding:
ACK 05ee33f05444b2bd1f37876676b7d471c5415710

PR #733, opened 2021-01-13
Edit 0.14.1 release notes to separate multiple PRs with "," instead of "/"

harding:
As mentioned today in #bitcoin-core-dev IRC, the release notes for 0.21.0 will include at least one reference to a PR that was merged in the new GUI repository.  laanwj said he'll put this in the notes as "gui/#nnnn", so this PR updates the plugin to handle that appropriately.

To eliminate conflicts, one historic release note is updated to change from using `/` as a separator between PRs to using `,` as the separator.  The comma is consistent with other release notes, such as that for 0.14.0.

This change should be forward compatible for other possible monotree repos as long as they're part of the bitcoin-core organization and the release notes use the format `<repo-name>/#<pr-number>`, e.g. `wallet/#123`.

### DIFF
--- a/_plugins/githubify.rb
+++ b/_plugins/githubify.rb
@@ -26,9 +26,17 @@ require 'yaml'
       ## Convert #1234 into URL for the pull request
       ## If #1234 links to an issue, GitHub automatically redirects
       #
-      ## Require at least two digits to reduce false positive matches
-      output.gsub!(/#([0-9][0-9][0-9]*)/){ |s|
-        '<a href="' + @repository_url + '/pull/' + $1 + '">' + s + '</a>'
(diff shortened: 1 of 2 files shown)


harding:
Tested:

- Compiled the current master (this PR's merge root) and this branch, then diff'd the output HTML of the two sites to confirm there were identical except for the expected `s/\//,/` change.

- Made a fake reference to a GUI PR:

laanwj:
Code review ACK 98b565a4f1d304e551547736477b4454c1ce2375

> To eliminate conflicts, one historic release note is updated to change from using / as a separator between PRs to using , as the separator. The comma is consistent with other release notes, such as that for 0.14.0.

Good catch. A good thing that separating with `,` is what we've mostly done.

PR #734, opened 2021-01-14
Releases: add 0.21.0

laanwj:

PR #735, opened 2021-01-14
Plugins: fix monotree repo ref from gui/#123 to gui#123

harding:
There was a slight miscommunication about the expected format in #733, this uses the format from #734 as tested on the actual release notes:

I also quickly checked the HTML diff from before and after this change and I only saw one unwanted change (a link to an issue was dropped).  I think that's acceptable to keep the pattern simple.


laanwj:
Right, I chose gui#12 because it's similar to github's bitcoin-core/gui#12

Sorry if I was confusing about it. Thanks for the update.

ACK 2ca4ef277d68ce195b93919e1112ef9e40b8b545

PR #741, opened 2021-01-20
Change all local links to be relative

raucao:
In order for the static website to work when hosted under nested paths/directories, e.g. `https://some-ipfs-gateway.com/ipfs/some-content-hash/`, all links and source file paths need to be relative to the respective documents, instead of absolute from the root.

(This branch may need some more adjustments before it can be merged, as I have not familiarized myself with the site release process yet.)

refs #739

### DIFF
--- a/_includes/root_path.html
+++ b/_includes/root_path.html
@@ -1,0 +1,18 @@
+{% assign ROOT_PATH = ''
+%}{% assign depth = page.url | split: '/' | size | minus: 1
+%}{% if depth < 1
+%}{% assign ROOT_PATH = '.'
+%}{% elsif depth == 1
+%}{% assign ROOT_PATH = '..'
+%}{% elsif depth == 2
+%}{% assign ROOT_PATH = '../..'
+%}{% elsif depth == 3
+%}{% assign ROOT_PATH = '../../..'
+%}{% elsif depth == 4
+%}{% assign ROOT_PATH = '../../../..'
+%}{% elsif depth == 5
+%}{% assign ROOT_PATH = '../../../../..'
+%}{% elsif depth == 6
+%}{% assign ROOT_PATH = '../../../../../..'
+%}{% endif
+%}
--- a/_includes/root_path.html
+++ b/_includes/root_path.html
@@ -1,6 +1,8 @@
 {% assign ROOT_PATH = ''
 %}{% assign depth = page.url | split: '/' | size | minus: 1
-%}{% if depth < 1
+%}{% if page.permalink == '/404.html'
+%}{% assign ROOT_PATH = ''
+%}{% elsif depth < 1
 %}{% assign ROOT_PATH = '.'
 %}{% elsif depth == 1
 %}{% assign ROOT_PATH = '..'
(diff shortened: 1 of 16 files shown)


_includes/root_path.html
```diff
@@ -0,0 +1,18 @@
+{% assign ROOT_PATH = ''
```
harding, line 1: This currently produces bad links when `assets/elements/404.md` is rendered as the content of that page is served at the URL of a non-existent page (e.g. https://bitcoincore.org/foo/bar ), and that URL may be more or less deep than the static 404 page.  I think maybe you could throw a `{% if page.permalink == "/404.html" %}{% assign ROOT_PATH="" %}{% endif %}` in there to fix, since I suspect IFPS doesn't have an automatic 404 redirect feature.

harding (changes requested):
A clever solution to creating relative links, I like it!  Lightly tested and it looks pretty good.  Currently our HTMLProofer test reports links on the 404 page and download page are broken.  The 404 one does need to be fixed; the download page one either needs the `--url-ignore` pattern in `Makefile` updated or some additional target/variable we can pass during build that's special for mirrors (including on IPFS).  Suggestions included inline.  Thanks @raucao !

_includes/root_path.html
raucao: Added the switch.

Since go-ipfs 0.7.0, the built-in HTTP gateway [supports custom 404 pages](https://github.com/ipfs/go-ipfs/blob/v0.7.0/CHANGELOG.md#custom-404). But I think you wouldn't configure a specific page for a single target/website, but a generic one for the gateway itself instead.

Not sure if this means the 404 links on bitcoincore.org would now be broken when looking at a nested path.

_includes/templates/download.html
```diff
@@ -6,7 +7,7 @@
 {% assign VERSION_SORTED_RELEASES = site.releases | sort: 'release' | reverse %}
 {% capture CURRENT_RELEASE %}{% for subver in VERSION_SORTED_RELEASES[0].release %}{{subver}}{% unless forloop.last %}.{% endunless %}{% endfor %}{% endcapture %}
 {% assign magnet = VERSION_SORTED_RELEASES[0].optional_magnetlink %}
-{% capture PATH_PREFIX %}/bin/bitcoin-core-{{CURRENT_RELEASE}}{% endcapture %}
+{% capture PATH_PREFIX %}{{ ROOT_PATH }}/bin/bitcoin-core-{{CURRENT_RELEASE}}{% endcapture %}
```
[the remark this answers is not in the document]
raucao: Just as additional input, it appears feasible to host binaries on IPFS, too: https://blog.ipfs.io/2020-02-14-improved-bitswap-for-container-distribution/

The question would be how to link them exactly, but going up one node relative to the root path should do the trick I guess.

kkqq2011 (changes requested):
Utilizing Ghost commander for file log {0000}. Need *0000 file to be reviewed for expression for D:CERT_PLUS_A:/%20%....
For transfer on MIT license directives.

MarcoFalke:
Needs rebase?

laanwj:
Thanks!

PR #745, opened 2021-01-21
RPC Docs: add 0.21.0

harding:
Generated using this repo's `contrib/doc-gen/generate.go`

Should be fully deterministic, so anyone wanting to review can install 0.21.0, put `regtest=1` in their `bitcoin.conf` file, start `bitcoind`, and run `go run generate.go`.

I lightly tested a local preview and didn't see anything broken this release.


harding:
Thanks @ben-kaufman for the ping in https://github.com/bitcoin-core/bitcoincore.org/pull/734#issuecomment-764692944 ; the OP above has the instructions for how to create the docs.

ben-kaufman (approved):
Thanks @harding, I checked myself and looks like I'm getting indeed the same result.

ben-kaufman:
Thanks for the explanation! I've made a PR adding these instructions to the relevant part of the Bitcoin Core release-process doc, so it would be easier to find how this update is done. Hope that makes sense to have there.

https://github.com/bitcoin/bitcoin/pull/20987

laanwj:
Cursory review ACK d63f30a5e0b110667c2868be168148f0142aa56c

PR #744, opened 2021-01-21
Revert "Remove bitcoin.pdf"

Legogris:
This reverts commit c131308cb9cf3538ba9db04a3002a76e6f334f62 in PR #740.

PR #740, opened 2021-01-20
Remove bitcoin.pdf

laanwj:
The licensing status of this file is unclear. It can be found in enough places by now, and as this site is about Bitcoin Core (a specific implementation), not about Bitcoin in general, the whitepaper doesn't necessarily need to be hosted here.

Unless anyone can point to an explicit place where Satoshi licensed it under a free license, legally it is safer to remove it from the bitcoin-core.org site.

PR #752, opened 2021-01-25
Revert "Remove bitcoin.pdf" take 2

x1ddos:
This reverts commit c131308cb9cf3538ba9db04a3002a76e6f334f62.

The original revert commit in https://github.com/bitcoin-core/bitcoincore.org/pull/744
was closed without any reasoning even though it articulated well
in regards to permissive licensing and why it is ok for bitcoin.pdf
to be served from bitcoincore.org.

The "Remove bitcoin.pdf" in https://github.com/bitcoin-core/bitcoincore.org/pull/740
did not provide any legal grounds for removing the PDF except for
"unclear status". In fact, it questions "unless anyone can point to an
explicit place where Satoshi licensed it under a free license" which
has been pointed out in https://github.com/bitcoin-core/bitcoincore.org/pull/744.

Unless there's a lawyer present who can refute or confirm the
reasoning, I see no basis for removal of bitcoin.pdf.


laanwj:
> Unless there's a lawyer present who can refute or confirm the reasoning, I see no basis for removal of bitcoin.pdf.

What we host on this site is our decision, not yours.

If you keep doing this, we will have to block you from the organization.

laanwj:
Look, sorry for me getting so annoyed here. It has been a wild few days. But I think a lot of people have a misunderstanding how free software / open source works.

It means "I give away this piece of code for people to use under a permissive license". That's all. You can copy it, add, modify or remove anything you want in your own copy. You can suggest changes to be merged upstream.

You are however not entitled to the author's work, or for anything to be included.

Reopening a new PR that is exactly like one that is closed is just jerk behaviour. You can host the whitepaper anywhere you want, or even a copy of bitcoincore.org with the whitepaper. But you can't bully us into hosting it if we don't want to.

PR #762, opened 2021-03-26
Downloads: fix URL

harding:
> There is currently no codesigning key available for Windows and to add insult to injury, Comodo revoked the old one. Until this can be resolved link the unsigned installer so that installation can at least continue with a big yellow warning.

> See bitcoin-core/gui#252.

Text above from #761 ; this is a simpler approach that updates all the links.  It can be reverted when the re-signed binary is uploaded.

In addition to this PR, I also checked that the shasums file is signed with @laanwj's usual key and that the hash of the -unsigned file is correct (I didn't check the other binaries).

### DIFF
--- a/_data/binaries.yaml
+++ b/_data/binaries.yaml
@@ -1,5 +1,5 @@
 win64zip: "win64.zip"
-win64exe: "win64-setup.exe"
+win64exe: "win64-setup-unsigned.exe"
 macdmg: "-osx.dmg"
 mactar: "osx64.tar.gz"
 lin64: "x86_64-linux-gnu.tar.gz"


harding:
The tests pass: https://travis-ci.org/github/harding/website/builds/764538282  Our testing does include the current download URLs.

MarcoFalke:
Keep in mind that the website hasn't been updated for months, so merging this won't change the download URL for users

laanwj:
Posthumous ACK

> Keep in mind that the website hasn't been updated for months, so merging this won't change the download URL for users\

Oh it is not updating again?

MarcoFalke:
Yeah, at least for me. https://bitcoincore.org/en/download/

PR #765, opened 2021-04-24
Rename the meeting summary page to fix duplicate names

azuchi:
Since the name of the meeting page and summary page are the same `meetings`, multiple english pages will be displayed when opened in other languages.

This PR fix duplicate names.

### DIFF
--- a/_posts/en/pages/2020-08-01-meeting-summaries.md
+++ b/_posts/en/pages/2020-08-01-meeting-summaries.md
@@ -3,7 +3,7 @@ type: pages
 layout: page
 lang: en
 title: IRC Meeting Summaries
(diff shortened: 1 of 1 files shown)


azuchi:
Screenshot from 2021-04-24 13-13-25

katesalazar:
Reproduced, and concept ACK.

katesalazar:
I think this is a bit confusing until a875f9072ee0ce2 is replayed by translators,
but I think this is good to go. Tested ACK.

harding:
Tested ACK d4c6d6efa586ddb92a1a466a3fbe08880052479e Thanks!

PR #796, opened 2021-09-14
Downloads: hide verification instructions

harding:
This one-character patch temporarily hides the verification instructions, which have become out of date with the release of 22.0 and combined signature files.  See controlling issue #793 

An update to the instructions will require bumping the page version number and so automatically show the instructions again.


harding:
Preview with this patch:

sipa:
Concept ACK (agree with hiding the instructions until they're updated).

prayank23:
ACK https://github.com/bitcoin-core/bitcoincore.org/pull/796/commits/6006441e1968ea2444714d637cbf9a1bad50b275

If anyone has issues with verification or looking for instructions can use the links mentioned in https://bitcoincore.org/en/contact/

laanwj:
ACK, this is a good idea until better instructions can be added.

PR #799, opened 2021-09-14
Update maintenance-table.md after 22.0 release

MarcoFalke:
* Remove 0.17 (EOL)
* Set EOL for 0.20
* Set release day for 22.0


_includes/posts/maintenance-table.md
```diff
@@ -1,6 +1,6 @@
-| 0.17 | 2018-10-03 | 2019-11-24  | 2020-08-01 |
 | 0.18 | 2019-05-02 | 2020-06-03  | 2021-02-01 |
 | 0.19 | 2019-11-24 | 2021-01-15  | 2021-08-01 |
-| 0.20 | 2020-06-03 | after v22.0 | TBA        |
+| 0.20 | 2020-06-03 | 2021-09-13  | 2022-02-01 |
```
harding, line 3: This is 5.5 months instead of the usual >6 months.  Is that intentional?

fanquake:
Changes look ok. Related https://github.com/bitcoin/bitcoin/issues/22969.

_includes/posts/maintenance-table.md
MarcoFalke: Happy to adjust by a few days, but I think the exact date doesn't matter much, as long as it gives the idea that the software can be considered unsupported after that date.

If a critical consensus bug was to be fixed on Feb 2nd (1 day after EOL), we might still backport a fix to the branch. If one is found a few months after EOL, less so.

MarcoFalke: It is not like someone sets an alarm on Feb 1st to unplug something. The date is mostly symbolic.

laanwj:
ACK d61f5edca6b7ed7920bd0c3aa3937b0b41697a6d

PR #804, opened 2021-09-18
Fix broken link in change log for 22.0 release note

azuchi:


harding:
Tested ACK 759c4c027d42da5a7db3ba7640ce1185f6be56db .  This implements the standard we used with the previous release.  If this change persists upstream, we can update our plugin.

Issue #811, opened 2021-10-04
ci: travis -> cirrus ci


MarcoFalke:
As travis was shut down, someone should translate the travis yaml to a cirrus ci yaml

laanwj:
Yes, it's not very… scalable for @harding to run the tests manually every PR :smile:

MarcoFalke:
See #812

PR #812, opened 2021-10-04
Replace travis by cirrus CI

laanwj:
Attempt at #811.

Loosely based on the [Ruby example](https://cirrus-ci.org/examples/#ruby) in the documentation, as well as the old Travis config.


.cirrus.yml
```diff
@@ -0,0 +1,16 @@
+container:
```
MarcoFalke, line 1: I think this is missing one (top-level) `task:`

laanwj: Yes it wasn't clear to me what should be nested under task, or if that was necessary. It seems "container" is outside of it at least here https://cirrus-ci.org/examples/#ruby

laanwj:
Locale issue?
```
htmlproofer 3.14.1 | Error:  invalid byte sequence in US-ASCII
make: *** [Makefile:15: test-slow] Error 1
```

laanwj:
Uh
```
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
```

laanwj:
That did it. New errors!
```
- ./_site/en/download/index.html
  *  4879:5: ERROR: Unexpected end tag : p (line 4879)
  *  4913:312: ERROR: Unexpected end tag : p (line 4913)
  *  4915:5: ERROR: Unexpected end tag : p (line 4915)
  *  4971:312: ERROR: Unexpected end tag : p (line 4971)
  *  4973:5: ERROR: Unexpected end tag : p (line 4973)
  *  5024:312: ERROR: Unexpected end tag : p (line 5024)
  *  5026:5: ERROR: Unexpected end tag : p (line 5026)
htmlproofer 3.14.1 | Error:  HTML-Proofer found 7 failures!
```

laanwj:
Ok, this should do it.
@harding can you take a look please, had to make some small changes to the site to make it pass the tests.

harding:
Tested ACK 90f4dbc8ae364af6749c661e03f88ebef7ede9ab .  Previewed the modified Download page; there's a slight change in rendering due to the removal of the `<pre>` tag.  I'm surprised that needed to be removed, but the difference is fine.

Ran all the tests locally and they passed (I knew they were failing on the `<p>` tags before and had meant to fix that).

Thanks, it'll be great to have CI running again!

laanwj:
>  I'm surprised that needed to be removed, but the difference is fine.

Same, figuring out why the hell it was failing took quite a while. But I think only inline HTML elements are allowed in `<p>…</p>`, not block elements?

laanwj:
OK, found the explanation:

> Paragraphs are block-level elements, and notably will automatically close if another block-level element is parsed before the closing \</p> tag. See “Tag omission” below.

— https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

So yes, while it isn't *forbidden* the `<pre>` will effectively close it.

PR #815, opened 2021-10-17
Replace Mac => macOS

katesalazar:
They call it macOS now, since a while.

If merged, this propagates to 808 (merged),
810 (unmerged) and 814 (unmerged),
among possibly others.


katesalazar:
0.21.2 is less than a month old. The corresponding web site release
page is current, not historical. I'd suggest reopening this, however
much more conveniently in the form of an issue.

PR #833, opened 2022-01-06
Link to guix.sigs repo rather than gitian sigs

fanquake:
Guix is now the release process, and linking to gitian is at least
causing confusion, see https://github.com/bitcoin-core/docs/issues/76.


MarcoFalke:
See also #807

harding:
Tested ACK 353e34ee5552e96a3ddcced7214d430ce7844132 Thanks!

PR #843, opened 2022-01-16
Add spanish translation of 0.20.2

katesalazar:


laanwj:
Thanks for adding this. It would be good if someone who knows Spanish could proof-read this and ACK it.

katesalazar:
np, my pleasure, dunno if there's any kind of **unintrusive**
way to peek translators about this area of work

PR #847, opened 2022-01-17
Fix broken links in the ZKCP article

katesalazar:
Transaction links are not actually broken, but soon to be.


kanzure:
ACK e7bdfef896fdbc194353be50d81bb0bf15009ed4

PR #858, opened 2022-02-01
Remove broken google search javascript

MarcoFalke:
The feature is broken for me (on Firefox and Chrome), so remove it. If someone wants to add it back later, I suggest switching to duckduckgo.

This removes another vector where the visitors IP is sent to google and remote javascript is executed on the visitors machine.


fanquake (commented):
ACK

harding:
Code review and quickly tested ACK d4fb3808ee7b2b7cd4d74473f091bb1144d50a52 Thanks!

PR #859, opened 2022-02-03
remove google plus

MarcoFalke:
Google plus was shut down long ago, so no need to keep any references to it. Also, remove other google settings.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -25,10 +25,7 @@ kramdown:
 # Site owner
 owner:
   name: Bitcoin Core
-  avatar: bio-photo.jpg
-  bio: ""
   twitter: bitcoincoreorg
-  github: bitcoin
 
 exclude:
   - _config.yml
--- a/_config.yml
+++ b/_config.yml
@@ -29,23 +29,7 @@ owner:
   bio: ""
   email:
   twitter: bitcoincoreorg
-  facebook: #username
   github: bitcoin
-  stackoverflow: #123456/username   from a "http://stackoverflow.com/users/123456/username" link
-  linkedin: #username
-  xing: #username
-  instagram: #username
-  lastfm: #username
-  tumblr: #username
-  pinterest: #username
-  foursquare: #username
-  steam: #username
-  dribbble: #username
-  youtube: #username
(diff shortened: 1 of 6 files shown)


_config.yml
```diff
@@ -30,13 +30,6 @@ owner:
...
-  google:
-    plus: #username
-    analytics: #UA-85008990-1
-    verify:
-    ad-client:
-    ad-slot:
-  bing-verify:
   github: bitcoin
   stackoverflow: #123456/username   from a "http://stackoverflow.com/users/123456/username" link
   linkedin: #username
```
laanwj, line 35: I guess if we're deleting unnecessary settings, there's quite a few below here as well which could go (all of them? I don't see a bitcoin core steam release happening any time soon…).

_config.yml
```diff
@@ -25,34 +25,7 @@ kramdown:
 # Site owner
 owner:
   name: Bitcoin Core
-  avatar: bio-photo.jpg
```
laanwj, line 28: Asset unused after this?

_data/authors.yml
```diff
@@ -2,40 +2,19 @@
```
laanwj, line 2: Is `authors.yml` still used at all?

MarcoFalke: Yes, for author.github, author.twitter and author.name. Example: https://bitcoincore.org/en/2015/12/14/segregated-witness/

I'll remove that in a follow-up.

_config.yml
MarcoFalke: Not yet. Still used in _includes/author-bio.html

I'll remove that in a follow-up.

_data/authors.yml
laanwj: Thanks!

MarcoFalke:
ci is green

katesalazar:
concept ACK, looking forward to review eventually

Zero-1729:
Concept ACK

katesalazar:
approach ACK, untested

MarcoFalke:
rfm? :)

PR #861, opened 2022-02-09
Final update release key expiration date

laanwj:
Change the expiration to 2024. The key is no longer being used for new releases, but letting it expire now would be inconvenient because there may still be another 0.21.x release, and, people may still want to verify older releases. I've also pushed this key to the `keys.openpgp.org` keyserver.

This should be enough for all releases signed with it to be EOL. After that, it can be removed from the site completely.


MarcoFalke:
Concept ACK

```
$ gpg --list-keys 01EA5486DE18A882D4C2684590C8019E36C2E964
pub   rsa4096 2015-06-24 [SC] [expires: 2024-02-09]
      01EA5486DE18A882D4C2684590C8019E36C2E964
uid           [ unknown] Wladimir J. van der Laan (Bitcoin Core binary release signing key) <laanwj@gmail.com>

**Comment by Zero-1729** [comment]
Concept ACK ca7a649c50ed244682b4e3b3237ec13d6a148640

```sh
$ gpg --import keys/laanwj-releases.asc 
$ gpg --list-keys 01EA5486DE18A882D4C2684590C8019E36C2E964
pub   rsa4096 2015-06-24 [SC] [expires: 2024-02-09]
      01EA5486DE18A882D4C2684590C8019E36C2E964
uid           [ unknown] Wladimir J. van der Laan (Bitcoin Core binary release signing key) <laanwj@gmail.com>
```

PR #869, opened 2022-04-25
Bitcoin Core 23.0

laanwj:
Add Bitcoin Core 23.0 release pages.


laanwj:
Re-pushed, the name of the `mactar` was incorrect. I've also made `macdmg` consistent in handling by not including `-` in `binaries.yaml`.

_data/binaries.yaml
```diff
@@ -1,7 +1,7 @@
 win64zip: "win64.zip"
 win64exe: "win64-setup.exe"
-macdmg: "-osx-signed.dmg"
-mactar: "osx64.tar.gz"
+macdmg: "x86_64-apple-darwin.dmg"
+mactar: "x86_64-apple-darwin.tar.gz"
```
laanwj, line 4: "mac" at some point should mean "x86_64" or "aarch64" but it would be good if someone who understands more about site design looks into this later

PR #873, opened 2022-04-28
contrib: Add fanquake to trusted keys

laanwj:
Add fanquake to the trusted keys. This will allow him to merge PRs to the website. The key id comes from bitcoin's trusted-keys.


MarcoFalke:
Maybe remove inactive keys that haven't been used in 2 years as well? Can be added back when there is indication that they will be used again.

laanwj:
That's much more dangerous (it can break the website update again because it will fail to validate the current commits) so I will leave that to a future change, possibly pending changes to the verify-commits scripts to allow specifying a last valid commit.

katesalazar:
ACK 4c61014da3bb438a01eeff609f987e317e6f751f

katesalazar:
ACK 4c61014da3bb438a01eeff609f987e317e6f751f

harding:
ACK 4c61014da3bb438a01eeff609f987e317e6f751f . Verified the same keyid appears in the bitcoin repository.  Thanks!

PR #878, opened 2022-06-10
adding more specific verification instructions

JayOceans:
need to verify fingerprints are authentic by comparing them in multiple places

### DIFF
--- a/_posts/en/pages/2017-01-01-download.md
+++ b/_posts/en/pages/2017-01-01-download.md
@@ -95,8 +95,7 @@ obtain_release_key: >
   key using this command:</p>
 
 choosing_builders: >
-  It is recommended that you choose a few individuals from this list who you find
-  trustworthy and import their keys as above, or import all the keys per the
+  It is recommended that you choose a few individuals from this list whose fingerprints you have confirmed to be authentic by comparing them on multiple trustworthy websites and import their keys as above, or import all the keys per the
   instructions in the <a href="$(BUILDER_KEYS_URL)"><code>contrib/builder-key</code>
   README</a>. You will later use their keys to check the signature attesting to the
   validity of the checksums you use to check the binaries.


asherhopp:
Makes sense to me, the instructions for verifying signatures are somewhat vague as-is.

katesalazar:
intended or not, this enshrines the for-profit social construct

NAK

mikeobank:
ACK

stickies-v (commented):
Concept ACK, but I would:
- add the extra trustworthiness verification instructions at the end of the paragraph, instead of at the beginning
  - rationale: these instructions are mostly aimed at people unfamiliar with the process, and the process can already be intimidating (or appear too much hassle) as is. Imo it's a big win already if people start out with just getting keys from `builder-keys`. Then, when that all worked, they _can_ (but if they don't, they're still much better off than not verifying at all) increase their trust in the keys by looking at other platforms etc). I can imagine that when some people read these extended instructions, they would just not verify at all because they don't know (or don't want to spend the time to investigate) what trustworthy websites are and therefore assume the whole thing is pointless. 
- in addition to looking at trustworthy websites, suggest comparing fingerprints with people they trust

_posts/en/pages/2017-01-01-download.md
```diff
@@ -95,8 +95,7 @@ obtain_release_key: >
   key using this command:</p>
 
 choosing_builders: >
-  It is recommended that you choose a few individuals from this list who you find
-  trustworthy and import their keys as above, or import all the keys per the
+  It is recommended that you choose a few individuals from this list whose fingerprints you have confirmed to be authentic by comparing them on multiple trustworthy websites and import their keys as above, or import all the keys per the
   instructions in the <a href="$(BUILDER_KEYS_URL)"><code>contrib/builder-key</code>
   README</a>. You will later use their keys to check the signature attesting to the
```
MarcoFalke, line 100: This file no longer exits. It might be better to directly link to the guix attestations repo and explain that each release is signed by a different set of builders. For example the 23.1 one is signed by https://github.com/bitcoin-core/guix.sigs/tree/main/23.1

MarcoFalke:
Are you still working on this?

darosior:
Another friendly ping, especially for updating the URL pointing to the builder keys as [it's confusing users](https://bitcoin.stackexchange.com/q/117125/101498).

PR #879, opened 2022-06-10
Update GPG4win path for easy user verification.html

VladdyC:
Updated the path to the Windows GPG installation to reflect the latest version of Gpg4win: "C:\Program Files (x86)\GnuPG\bin\gpg.exe"


laanwj:
ACK, i've checked this with "gpg4win-4.0.2.exe" (the latest version at this point, see https://gpg4win.org/download.html) in WINE and it installed the `gpg.exe` in the same path.

Issue #868, opened 2022-04-21
v23 rpc docs are missing


katesalazar:
v23 is tagged
https://github.com/bitcoin/bitcoin/releases/tag/v23.0

laanwj:
See #869, the site doesn't seem to update though dunno why.

katesalazar:
> ~See #869~
> 
> ~Edit: oh this used to be about the site, I'm not sure it's a good idea to recycle issues for different topics, in any case the site is updated now.~

@laanwj sorry for the mess and thank you very much for getting the site up to date.

PR #887, opened 2022-08-03
RPC Docs: add 23.0.0

andrewtoth:
Generated using this repo's `contrib/doc-gen/generate.go`.

Should be fully deterministic, so anyone wanting to review can install 23.0.0 and run the following:
```
bitcoind --daemon --regtest
cd contrib/doc-gen
go run generate.go
```

I lightly tested a local preview and didn't see anything broken this release.

Closes https://github.com/bitcoin-core/bitcoincore.org/issues/868.


stickies-v:
Ran a diff with my locally generated docs, looks like you're missing the `rpc/zmq/` subdir, so I think maybe you've configured with `--without-zmq`? Beyond that, everything matches and I also compared with the [changelogs](https://bitcoincore.org/en/releases/23.0/) (not every single change, but quite a bit of it) and that all seems to align. So, once `rpc/zmq/` is added it's an ACK from me.

andrewtoth:
@stickies-v good catch thanks. I reran with a binary from bitcoincore.org and it added `zmq/getzmqnotifications.html`.

stickies-v (approved):
ACK 62ef4aecf836e5f56b69bde64d66baf95542cc9d

Matches my docs and seems to incorporate all changes (lightly tested) in the [changelogs](https://bitcoincore.org/en/releases/23.0/). Thanks for updating this!

PR #892, opened 2022-08-16
Add `--ignore-missing` flag to `shasum` command on MacOS

mikeobank:
Shasum version 6 on MacOS also supports the `--ignore-missing` flag like on Linux. Allows for a cleaner user experience. 

**Question:**
Is there a some form of linting to define and take care of the whitespace stripping done on line 22 - 24? Or should those changes be reverted?

**To Do:**
The instructions for MacOS and Linux (https://bitcoincore.org/en/download/) still state: *In the output produced by the above command, you can safely ignore any warnings and failures, but you must ensure the output lists "OK" after the name of the release file you downloaded.* Think *you can safely ignore any warnings and failures* could be removed in both cases. And maybe add a warning that the output can be empty, in the case the user has no download at all in the current working directory.

### DIFF
--- a/_includes/templates/download.html
+++ b/_includes/templates/download.html
@@ -19,9 +19,9 @@
 {% capture BUILDER_KEYS_TXT_URL %}{{page.builder_keys_url}}/keys.txt{% endcapture %}
 
 {% capture OBTAIN_RELEASE_KEY %}
-  {{page.obtain_release_key | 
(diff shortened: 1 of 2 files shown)


stickies-v:
Concept ACK - would undo the whitespace changes though, not relevant to this PR.

I also use `--ignore-missing` on macOS and imo that's a better approach than to suggest ignoring all warnings and failures.

fanquake:
Picked this up in #1027.

PR #907, opened 2022-10-20
Adding pt-BR/rss

mateusnds:
When translating pages into Portuguese, I get an error saying that there is no rss in Portuguese, so I'm trying to create to follow the page translations

https://github.com/bitcoin-core/bitcoincore.org/pull/905 https://github.com/bitcoin-core/bitcoincore.org/pull/906

Ref:
***
  *  internally linking to /pt-BR/rss/, which does not exist (line 197) <a href="/pt-BR/rss/">RSS rss feeds</a> htmlproofer 3.14.1 | Error:  HTML-Proofer found 1 failure! make: *** [Makefile:15: test-slow] Error 1
***

Use this page to open Pull requests for changes to the Bitcoin Core project
website at [bitcoincore.org](https://bitcoincore.org).

[bitcoin.org](https://bitcoin.org) is not connected with or maintained by the
Bitcoin Core project contributors. Pull requests for changes to that website
should be opened at [the bitcoin.org
repository](https://github.com/bitcoin-dot-org/bitcoin.org/pulls).

General bitcoin questions and/or support requests are best directed to the
Bitcoin StackExchange at https://bitcoin.stackexchange.com.

PR #905, opened 2022-10-18
Add portugues transalation to ABOUT page

mateusnds:
I'm starting to translate the site into Portuguese, this is the first page I'm doing and I hope everything goes well-. If I'm doing something wrong, please let me know.

Use this page to open Pull requests for changes to the Bitcoin Core project
website at [bitcoincore.org](https://bitcoincore.org).

PR #908, opened 2022-10-20
Index Portugues page

mateusnds:
When I tried to create the RSS page, it gave an error that there is no index.

Ref:
https://github.com/bitcoin-core/bitcoincore.org/pull/907


fanquake:
Please consolidate your changes into a single PR. i.e #905. Rather than opening many different ones.

PR #914, opened 2022-11-14
rpc docs: remove docs for EOL versions

fanquake:
Removes the RPC documentation for 0.16.0, 0.16.1, 0.16.2, 0.16.3, 0.17.0, 0.18.0, 0.19.0, 0.20.0.
Keeps 0.21 for now as it's only just gone EOL (happy to remove if preferred).
24.0 should be added shortly.


vincenzopalazzo (commented):
maybe here instead of removing the doc that it is EOL it is better to add an alert in the index that it is time to upgrade?

glozow:
Would it be difficult to keep these? Agree cleanup and reducing the amount of things to maintain is good, but I must admit that there are a few links to RPC docs in the review club website, e.g. [here](https://bitcoincore.reviews/25991).

fanquake:
Ok. Will leave these as-is for now. Maybe we should add some/all of the missing versions?

0xB10C:
Issue about potentially adding docs for older versions https://github.com/bitcoin-core/bitcoincore.org/issues/921#issue-1448836204

PR #918, opened 2022-11-14
Invert order of maintenance-table.md

stickies-v:
It's a bit of a nit, but I find it a bit counterintuitive to put the oldest release at the top. Releases become less important the older they get, so this makes more sense to me.


0xB10C:
±0 on this. Checked that the table rows are now reversed.

katesalazar:
I'd say rows are currently sorted in descending urgency.

It's gonna take many more paragraphs to convince me to ACK.

fanquake:
~0

Issue #917, opened 2022-11-14
arm64 darwin missing from download page


fanquake:
arm64 darwin [binaries are available as part of the 23.0 release](https://bitcoincore.org/bin/bitcoin-core-23.0/), however they are not listed on https://bitcoincore.org/en/download/ at all.

katesalazar:
Did anyone mention any plan for publishing dual architecture dmg?
(I suppose that's a thing...)

jarolrod:
@katesalazar why provide a fat binary? Lean binary for each architecture is preferable in my opinion.

Issue #696, opened 2020-02-06
Auto-detect non-x86 Linux systems on Download page


harding:
The big blue button on the download page uses the browser user agent string to recommend a binary for the user to download.  Presently it only recommends x86 binaries for Linux users.  Either it should also recommend ARM and RISCV binaries for those users or we should just get rid of the button and so encourage users to select their own platform.

![2020-02-06-09_06_51_451502339](https://user-images.githubusercontent.com/61096/73949705-ca206080-48c8-11ea-8d46-17b7b2772475.png)

List of popular browser strings to test matching against: https://www.whatismybrowser.com/detect/what-is-my-user-agent

Help needed: I don't see any browser strings on the above list for RISCV.  Does X11 run on that platform?  If so, anyone running a graphical browser could help by replying with their browser string (e.g. as reported [here](https://www.whatismybrowser.com/detect/what-is-my-user-agent)).

Previous discussion: https://github.com/bitcoin-core/bitcoincore.org/pull/695#pullrequestreview-354439276

katesalazar:
> Presently it only recommends x86 binaries for Linux users. Either it should also recommend ARM and RISCV binaries for those users or we should just get rid of the button and so encourage users to select their own platform.

I think it should recommend all of x86, ARM, RISCV, whatnot, Linux binaries (maybe ordered by "common sense", dunno), until you find the way to reliably detect the computer type/architecture. I would suggest not removing the big blue button. I think there can't be three links in the button, so the hell with it and add as many big blue buttons as architectures.

The architecture problem hitting Apple as well is imminent, I suppose.

---

BTW my resolution totally wrecks the big blue button off.

Typical:
<img width="824" alt="Screen Shot 2022-04-19 at 23 39 36" src="https://user-images.githubusercontent.com/52637275/164106421-7158768b-9a9e-4720-b268-948a2a3f73b4.png">

Artificially enlarged:
<img width="1031" alt="Screen Shot 2022-04-19 at 23 40 42" src="https://user-images.githubusercontent.com/52637275/164106525-012f5878-a038-46c5-a269-aa34ef73707e.png">

Will try to tweak and if get anything interesting add ticket and PR.

PR #920, opened 2022-11-14
download: add links for arm darwin binaries

katesalazar:
This shall close ticket #917.


_includes/templates/download.html
```diff
@@ -48,11 +48,19 @@ <h2>{{ page.latestversion }} {{CURRENT_RELEASE}} <a type="application/rss+xml" h
...
         <img src="/assets/images/os/med_osx.png" alt="osx">
         <span>
-          <a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX }}-{{ site.data.binaries.macdmg }}">macOS</a>
+          <a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX }}-{{ site.data.binaries.macdmg }}">macOS (x86_64)</a>
           <span><a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX }}-{{ site.data.binaries.macdmg }}" class="dl" id="macdmg">dmg</a> -
             <a href="{{ PATH_PREFIX }}/{{ FILE_PREFIX }}-{{ site.data.binaries.mactar }}" class="dl" id="mactar">tar.gz</a></span>
         </span>
       </div>
+      <div>
+        <img src="/assets/images/os/med_osx.png" alt="osx" style="visibility:hidden">
```
fanquake, line 57: Why `visibility:hidden`?

fanquake (commented):
Can you provide a screenshot of what this actually looks like.

_includes/templates/download.html
katesalazar: Intended to make invisible the image while having the image occupy the same block of space.

Although DOM-untidy, this tends to render OK-ish.

katesalazar:
Screen Shot 2022-11-15 at 10 56 42
Screen Shot 2022-11-15 at 10 56 24

_includes/templates/download.html
MarcoFalke: Agree  it should be visible if it takes up another square

katesalazar: force pushed a rev, same contents, just removing the visibility attr

fanquake:
Can you provide a new screenshot. Either by updating your previous comment, or deleting it and posting a new one.

katesalazar:
screenshots using d83eeeb77ceaef62a196e5263177b51ca13e2971
my normal resolution:
Screen Shot 2022-11-15 at 14 28 08
if I maximize:
Screen Shot 2022-11-15 at 14 28 21

fanquake (approved):
I think this looks fine now. Haven't tested.

hebasto:
Should this snippet https://github.com/bitcoin-core/bitcoincore.org/blob/d83eeeb77ceaef62a196e5263177b51ca13e2971/_includes/templates/download.html#L321-L324 be also modified to assign the correct file to the Download button when browsing on macOS arm64?

katesalazar:
ah, the tricky bit

I can push another rev with some swiftly changes,
but I don't have the devices to check if it'll be working

fanquake:
There's no need to bundle that change into this PR. It can be done separately as part of #696.

katesalazar:
I think a68a7e89b76 should do, but I have not the hardware to check

if does look good write something and I squash asap

_includes/templates/download.html
```diff
@@ -295,7 +295,12 @@ <h2 style="text-align: center">{{page.build_reproduction}}</h2>
 
 <script type="text/javascript">
 var os = 'windows64';
-if (navigator.userAgent.indexOf('Mac') != -1) var os = 'mac'
+if (navigator.userAgent.indexOf('Mac') != -1) {
+	var os = 'mac';
+	if (navigator.userAgent.indexOf('Intel') == -1) {
```
katesalazar, line 300: yeah... I'm reading stuff that says this doesn't work

katesalazar: at least this rev doesn't break anything for the intel macs

fanquake:
> There's no need to bundle that change into this PR. It can be done separately as part of https://github.com/bitcoin-core/bitcoincore.org/issues/696.

Can you just drop the untested changes.

katesalazar:
> > There's no need to bundle that change into this PR. It can be done separately as part of #696.
> 
> Can you just drop the untested changes.

Yeah, done!

PR #922, opened 2022-11-15
Maintenance table: highlight past dates

katesalazar:
Insist against ticket 918.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 _site
 .sass-cache
 .jekyll-metadata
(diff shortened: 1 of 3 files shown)


fanquake:
NACK - I don't think this is an improvement, and seems like overkill. I don't think we need to introduce any more javascript.

fanquake:
Going to close this for now. #918 has also been closed.

katesalazar:
fair - this change works, but it is dirty as rushed

PR #931, opened 2022-11-21
schedule: update maintenance table

fanquake:
Update schedule for the release of 24.0.
Rough schedule for v25.0 is now available: https://github.com/bitcoin/bitcoin/issues/26549.


katesalazar:
> Should we drop 0.20 now as well?

I think so. It can come back so easily if anyone asks.

MarcoFalke (approved):

PR #943, opened 2022-12-21
ci: Schedule PRs faster

MarcoFalke:
Avoids waiting


MarcoFalke:
Not sure why the credits thingy doesn't work for me, but reducing the CPU/memory should already improve waiting

fanquake:
Closing for now, given the CI is not running in any case. Could re-open once we've made some infra updates.

PR #951, opened 2023-02-07
keys: Remove laanwj-releases key

laanwj:
Remove laanwj-releases key. This key was used to sign the old gitian release binaries 0.21.x and before (which were pgp-signed using only one key), and is not going to be used again. It will expire in 2024 but I'm probably going to revoke it sooner.

There are no references to it in the text of the site.


MarcoFalke:
lgtm

0.21 is EOL for more than 4 months. Makes sense to remove the unused key.

PR #952, opened 2023-02-07
Remove laanwj from trusted-keys

laanwj:
allow-revsig-commits generated using

    git log --format="%H %GK" --merges $(cat contrib/verify-commits/trusted-git-root)..master | grep -E " 1E4AED62986CD25D" | cut -c -40 >> contrib/verify-commits/allow-revsig-commits

(thanks ajtowns for the improved script)


MarcoFalke:
I think you can just bump the trusted git root, like it was done before?

I never understood the difference, and at least it works in practise.

fanquake:
> I think you can just bump the trusted git root, like it was done before?

@MarcoFalke Want to open a PR to do this?

fanquake:
> but please check.

Looks correct to me. Going to go-ahead and merge this.

PR #826, opened 2021-12-17
Bump Ruby to 2.7.4, update github-pages and html-proofer gems

Sjors:
This requires updating the server to Debian Stable Bullseye, cc TheBlueMatt.

I tested a few pages locally and didn't notice any difference.

I also ran the `htmlproofer` incantation from the README, which I updated to match `Makefile`. Only one minor change was needed to satisfy the updated version.

Closes #825

PR #958, opened 2023-03-27
deps: Ruby 2.7.4

pinheadmz:
2.7.4 is the latest ruby version in debian stable according to https://packages.debian.org/stable/ruby/ruby

upgraded deps but kept `htmlproofer` on major version 3 because of: https://github.com/gjtorikian/html-proofer/issues/722

Opening as draft now to test against CI


fanquake:
Note that bumping this would likely also require changes to server infra.

fanquake:
Actually, dupe of #826,

pinheadmz:
🤦

PR #963, opened 2023-05-05
Minimize dependencies and add CI that matches the build server

achow101:
We don't need all of the github pages dependencies, just jekyll and a few others.

Also add a CI task that behaves similarly to the site's build server so we can catch issues earlier.


glozow (commented):
ACK 59485146e81cdde48f0069a16619849a2f1fdfb9

built fine locally. CI is green and the yaml looks fine.

PR #966, opened 2023-05-11
remove outdated calendar links

josibake:


glozow (commented):
@achow did you say you pinged about updating the calendar event time? concept ACK to removing incorrect time but could be kept if changed.

achow101:
The linked calendar appears to be updated (with the exception of the wallet meeting), so I don' think this matters?

glozow:
Nice! Close?

PR #973, opened 2023-05-26
Add japanese translation for 24.1

azuchi:


fanquake:
Thanks. 25.0 is now also available for translation.

PR #986, opened 2023-09-01
contrib: Fix commit history for verify-commits.sh

achow101:
`verify-commits.sh` currently fails as a mistake was made when merging a few PRs earlier this year that resulted in the merge commits being merged incorrectly and signed with the wrong key. To resolve this, a merge commit signed with a trusted key is added which has a parent of the last verify-commits-passing commit, and the current HEAD. This ensures that there is a path of commits which are signed by trusted keys thus resolving the verify-commits issue.


achow101:
> You should be able to fix this by creating a new merge commit which has a left branch of the last verify-commits-passing commit, and a right branch of the current tip.

Done.

achow101:
Note: Merging this with the merge script does not resolve the verify-commits issue because the resulting merge commit has the current tip as the left parent.

PR #1011, opened 2024-04-01
stats: update datadir_gb & pruned_gb

### LINKED ISSUE #1010: Storage required considerably more than 6GB now.
"Bitcoin Core requires a one-time download of about 500GB of data plus a further 5-10GB per month. By default, you will need to store all of that data, but if you enable pruning, you can store as little as 6GB total without sacrificing any security."

The download is closer to 600GB now. But more importantly:

The chainstate is now 10 GB and the lowest GUI prune setting is 1 GB so we're off by 5 GB almost 100% there.
https://github.com/bitcoin/bitcoin/blob/61de64df6790077857faba84796bb874b59c5d15/src/kernel/chainparams.cpp#L122

fanquake:
Should close #1010.

### DIFF
--- a/_data/stats.yml
+++ b/_data/stats.yml
@@ -7,9 +7,9 @@ datadir_gb: "600"
 ## Suggest rounding up a little bit to give room for growth in the UTXO
 ## set and to account for things like pruning only happening
 ## at blk file boundaries and mempool saving.
-pruned_gb: "6"
+pruned_gb: "10"
 
-## Size of monthly increase in dadadir size.  For 1 MB blocks, 144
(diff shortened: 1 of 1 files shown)


MarnixCroes (approved):
lgtm b2fccc15f0f876c06a9d3d4f058a610d21a5609e

glozow (commented):
Matches https://github.com/bitcoin/bitcoin/blob/6dabb315c4fb2cdc4421ed34dee5e19042f899c3/src/kernel/chainparams.cpp#L121-L122, so ACK b2fccc15f0f876c06a9d3d4f058a610d21a5609e

PR #1016, opened 2024-04-16
ci: Use bundler 2.4.22

achow101:
The last version of bundle to support ruby 2.7.4 is 2.4.22. Should fix the current CI failure.


laanwj:
> Should fix the current CI failure.

Looks like it does!

ACK 0c10f283febf3ced1e762de1df31f2d981d2bed7

PR #1026, opened 2024-06-07
maintenance: remove EOL 24.0 & use .x over .0

fanquake:
Using `.x` should make it clearer that it is the branch that is maintained. In particular, the most-recent point release on the branch.


glozow (commented):
lgtm 65829aa8c11f200b481dda22131f9abe1f7df0af

just to bikeshed a little bit, * might be more intuitive than x

fanquake:
I've used `.x` because that is what we currently use for branches: https://github.com/bitcoin/bitcoin/branches.

glozow:
Yes, but the website documentation is meant for users of the software, not developers of the software. I don't feel strongly though of course.

darosior:
Concept ACK.

dergoegge (approved):

glozow:
Merging, main point is to remove eol 24. Sounds like we will have more opportunities to make lifecycle info clearer in the near future

darosior:
> "Only the last point release of each supported major release version will have the latest security fixes."

I'm not sure this wouldn't mislead users into believing minor releases would always have the latest security fixes, which we unfortunately can't guarantee.

PR #1028, opened 2024-06-12
download: default to ARM macOS downloads over x86_64

fanquake:
If a mac is detected by the browser, default to the ARM macOS download, rather than x86_64. This seems more reasonable at this point. In future, we might be able to use more advanced browser logic to detect the actual architecture dynamically. i.e https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues.

### DIFF
--- a/_includes/templates/download.html
+++ b/_includes/templates/download.html
@@ -311,8 +311,7 @@ var linkwinexe = document.getElementById('downloadwinexe');
 var linkwinzip = document.getElementById('downloadwinzip');
 var hrefwin64exe = document.getElementById('win64exe').href;
 var hrefwin64zip = document.getElementById('win64zip').href;
-var hrefmaczip = document.getElementById('maczip').href;
-var hrefmactar = document.getElementById('mactar').href;
+var hrefmaczip = document.getElementById('macarmzip').href;
 var hreflin64 = document.getElementById('lin64').href;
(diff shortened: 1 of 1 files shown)


willcl-ark (approved):
crACK 96da1ff1b61b7d03d115825a8864566f12ffc6b1

I'm no web-expert, but the rationale of this change seems correct to me.

Apple has not sold new x86 Macs for a while now; the Mac Mini was the final hold-out -- all of which are now [Apple M* chip-based](https://www.apple.com/mac-mini/specs/).

Additionally, as I think alluded to in description, parsing the user agent will not be able to directly detect ARM vs x86 (for backwards-compatibility reasons, e.g. my M3 macbook running 14.4.1 has user agent `"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:126.0) Gecko/20100101 Firefox/126.0"` on firefox).

With that in mind it makes sense to update the default download to ARM, ~4 years after the introduction of the new chip and now the entire product line has migrated.

The only argument against that I can think of is that technically, ARM macs can run x86 apps via Rosetta, but not the other way around (AFAIK). However, I think a user who encounters an application startup error should be capable of reading the error and returning to bitcoincore.org and selecting the `macOS (x86_64)` download.

PR #1029, opened 2024-06-17
Update 2017-01-01-download.md

adminseedlatam:
In the link of https://bitcoincore.org/es/download/ where you have the different instructions to verify the links redirect to page that doesn't exist anymore.

The link of "builder_keys_url" is not `https://github.com/bitcoin/bitcoin/tree/master/contrib/builder-keys` the supposed to redirect. 

The correct value of builder_keys_url is: `https://github.com/bitcoin-core/guix.sigs/tree/main/builder-keys`

Also, the redirection of `fanquake.gpg` is `https://github.com/bitcoin-core/guix.sigs/blob/main/builder-keys/fanquake.gpg`


fanquake (approved):
ACK 53acc50b3f458e7467c6309fcb2a498f78ac85c9

PR #1032, opened 2024-06-25
release: Update 27.1 magnet url

achow101:
Since the SHA256SUMS.asc had to be reuploaded, the torrent has also changed. The magnet url should be updated to match the torrent file.


fanquake (approved):
ACK fe698a189395d4692546b2a4197074669cb08396

PR #1033, opened 2024-06-26
pages: add Security Advisories

fanquake:
Add a new page summarizing the upcoming security advisory policy.

advisories

### DIFF
--- a/_posts/en/pages/2024-06-26-disclosures.md
+++ b/_posts/en/pages/2024-06-26-disclosures.md
@@ -1,0 +1,50 @@
+---
+title: Security Advisories
+name: Security Advisories
+id: en-security-advisories
+permalink: /en/security-advisories/
+layout: page
+type: pages
+lang: en
+version: 1
+---
+{% include toc.html %}
+
+This page summarizes policies in relation to disclosing vulnerabilities in Bitcoin Core,
+as well as provides a summary of historical Security Advisories.
+
+## Policy
+
+When reported, a vulnerability will be assigned a severity category. We
+differentiate between 4 classes of vulnerabilities:
+
+* **Low**: bugs which are hard to exploit or have a low impact. For instance
+  a wallet bug which requires access to the victim's machine.
+
+* **Medium**: bugs with limited impact. For instance a local network remote crash.
+
+* **High**: bugs with significant impact. For instance a remote crash, or a local network RCE.
+
+* **Critical**: bugs which threaten the whole network's integrity. For instance an inflation or coin theft bug.
+
+**Low** severity bugs will be disclosed 2 weeks after a fixed version is released.
+  A pre-announcement will be made at the same time as the release.
+
+**Medium** and **High** severity bugs will be disclosed 2 weeks after the [last
+  affected release goes EOL](/en/lifecycle/). This is a year after a fixed version was first
+  released. A pre-announcement will be made 2 weeks prior to disclosure.
+
+**Critical** bugs are not considered in the standard policy, as they would most likely require an
+  ad-hoc procedure. Also, a bug may not be considered a vulnerability at all. Any reported issue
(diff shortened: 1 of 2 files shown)


kanzure:
Why is a policy needed? What does the policy actually do or what is the intention of having this policy on the site?

darosior:
Because a standard process is preferable to arbitrary handling. For all vulnerabilities but the most critical ones (which would most likely need to be dealt on a case by case basis) it's possible to follow a standardized disclosure process.

darosior:
See for instance https://www.erisian.com.au/bitcoin-core-dev/log-2024-06-06.html for previous discussions about this.

fjahr:
ACK 811d10d0564ef878cab5bb65682a10f9507925ef

The text looks good to me and I agree with the policy itself. I have mostly followed the discussion in IRC and below the gist draft and I think this reflects the public consensus.

kanzure:
Anyone on the security team can just disclose on this timeline, without a written policy. What is the purpose of this policy?

_posts/en/pages/2024-06-26-disclosures.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Disclosures
```
dergoegge, line 2: nit: I think "Security Advisories" would be the better title since it is widely established terminology:

* https://www.mozilla.org/en-US/security/advisories/
* https://www.vmware.com/security/advisories.html
* https://sec.cloudapps.cisco.com/security/center/publicationListing.x
* https://www.dell.com/support/security/en-ca

kanzure:
After talking privately with a few people, a reason for this was disclosed to me that makes more sense than the other reasons written in the previous comments. My understanding is that a few people want this reason to remain private.

I still think it makes more sense to call it a "procedure", "intention", "convention", or "process" long before "policy", and if there is a need for "policy" then make a Bitcoin Core Organization Policy LLC holding entity similar to the certificates silo entity....

As to the current proposed document itself.... what about issues that require changes to bitcoin consensus vs those that do not? like what if it's a low/medium but a consensus bug?

if I had been asked what ought to be in a disclosure "policy", then my expectation would have been a plea to security researchers to follow thoughtful and responsible steps because such a large amount of people's money is directly entangled with this project moreso than the rest of the software ecosystem. There have been many discussions on IRC, the mailing list, reddit, twitter/X as to what responsible disclosure looks like in the world of bitcoin that can also be referenced or drawn upon.

darosior:
I'm not really interested in bikeshedding how this should be named. If a number of contributors to the project believe it should be named "process" instead of "policy" i'm fine with it.

As for the comment about what the policy should be, feel free to suggest any concrete improvement. This has been discussed off and on for the past 9 months, and more actively in the past 3 months, by active contributors to the project. As far as i can tell, the vast majority of active Bitcoin Core contributors believe following this process to track and publicly disclose vulnerabilities is better than the status quo. I don't think this improvement should get blocked on a few vague mentions to past discussions which you seem to imply we would not even be aware of.

kanzure:
> [...] As far as i can tell, the vast majority of active Bitcoin Core contributors believe following this process to track and publicly disclose vulnerabilities is better than the status quo. I don't think this improvement should get blocked [...]

Nothing about my comments are able or intended to stop anyone from tracking or disclosing vulnerabilities. Thank you for understanding my messages.

_posts/en/pages/2024-06-26-disclosures.md
fanquake: Changed.

dergoegge (approved):
ACK 593e7406ca14c6b8169ab60cecf3f16e09d86b15

darosior (commented):
ACK 593e7406ca14c6b8169ab60cecf3f16e09d86b15

PR #1036, opened 2024-06-27
Backfill lifecycle table

murchandamus:
- Reverse order to put newest version at the top
- Put "after version MAJOR.0" for all maintenance end and EOL dates
- Backfill all dates up to version 0.8.x from file’s history in repository


murchandamus:
I was surprised that https://github.com/bitcoin-core/bitcoincore.org/pull/1026 removed the entry for the latest Major branch that went EOL. I would suggest to order the releases newest first, and I backfilled all the versions that we have data for from the prior versions of this repository.

I also amended the description of the lifecycle to explain the abbreviation EOL on its first appearance rather than later in the text.

stickies-v (commented):
I suggested reversing the order in #918 but there was not much enthusiasm for it - I remain supportive of the change even if I don't care too much either way.

With regards to the full history of versions, I think that would be nice to list on the website in full. I'm not sure if Lifecycle is the best place for it since the list is already quite long and will only ever expand, but again, not a huge concern at the moment either.

So overall:
- light +1 on reversing the order
- light +1 on backfilling the history, although I wonder if there's a better place for it (without duplicating)
- +1 on the EOL and TBA improvements

achow101:
ACK 0b15471cd8a968880a3788587bf356083443ebfe

I think it's a good practice to keep a listing of when each version became EOL. Similar listings exist for [debian](https://wiki.debian.org/DebianReleases) and [ubuntu](https://wiki.ubuntu.com/Releases).

darosior:
Concept ACK.

PR #1038, opened 2024-06-29
Fix default twitter/og meta tag image (fixes #1037)

### LINKED ISSUE #1037: Fallback image for Twitter/OG meta tags point to non-existent image [open-graph.html]
Currently, the page head rendering is pointing to a non-existent image for Twitter/OG image meta tags (open-graph.html fallback logic):

```
<meta name="twitter:image" content="https://bitcoincore.org/assets/images/default-thumb.png">

<meta property="og:image" content="https://bitcoincore.org/assets/images/default-thumb.png">
```

Should we point to one of the existing images or bring a new one for this? bitcoin_core_logo.png might be a good candidate.

joetor5:


achow101:
Do you have an example what the before and after of this change looks like?

fanquake:
> Do you have an example what the before and after of this change looks like?

The before is:
before

So I think pointing to an image that exists will be an improvement.

joetor5:
I created a temporary test post on my personal site with the same image (unmodified) and the twitter/og image meta to see how it would look like after. I tried using ngrok with my local bitcoincore site to test but the image was not displaying/replaced with other ngrok stuff when doing a preview.

Link: https://joetor5.github.io/bitcoin/2024/08/01/bitcoin-og-test.html

Preview on Twitter:

Screenshot 2024-08-01 at 9 34 37 AM


Maybe too big?

It is more problematic if the Twitter theme is in dark mode (you'll have to hover it to see something, but barely visible):

Screenshot 2024-08-01 at 9 39 49 AM

fanquake:
In that case, maybe it's actually just best to leave it as-is for now.

joetor5:
Agree.

PR #1041, opened 2024-07-02
deps: update ruby and dependencies for debian bookworm

achow101:
The build server updated to debian bookworm which has ruby 3.1.2, so update everything to require that. Also updates the other dependencies in the Gemfile.lock.


fanquake:
https://cirrus-ci.com/task/5661091059990528?logs=build_and_test#L87:
```bash
bundler: failed to load command: htmlproofer (/usr/local/bundle/bin/htmlproofer)
/usr/local/bundle/gems/html-proofer-5.0.9/lib/html_proofer/configuration.rb:68:in `parse_cli_options': invalid option: --check-html (OptionParser::InvalidOption)
```

I think the CI check can be changed to something like `bundle exec htmlproofer --disable-external --checks Links,Images,Scripts --ignore-urls '/^\/bin/.*/' --no-enforce-https ./_site`. That seems to work locally:
```bash
bundle exec htmlproofer --disable-external --checks Links,Images,Scripts --ignore-urls '/^\/bin/.*/' --no-enforce-https ./_site

Running 3 checks (Images, Links, Scripts) in ["./_site"] on *.html files ...


Checking 3085 internal links
Checking internal link hashes in 151 files
Ran on 2383 files!


For the Links check, the following failures were found:

* At ./_site/logs/2016-05-zurich-meeting-notes.html:505:

  'a' tag is missing a reference

HTML-Proofer found 1 failure!
```
Which results in a single failure.

achow101:
Yeah, working on the CI failure. I think it would be reasonable to keep https enforcement as all of our links should work with https anyways (checking that now).

achow101:
On second thought, there's a bunch of dead http links and a few sites that don't have https, so going with `--no-enforce-https`. Also fixed the `a` tag problem.

achow101:
Apparently debian only has html-proofer 3.19.2, so I set the Gemfile to force a 3.x version rather than the latest. This also means we won't have to change the flags passed to it.

PR #1046, opened 2024-07-05
Add japanese translations for security advisories

azuchi:


fanquake:
Unrelated, but something else you might also want to translate in a new PR?

text

PR #1050, opened 2024-08-01
Add japanese translations for security advisories in Bitcoin Core v22.0

azuchi:


fanquake:
Was going to say you might want to wait for #1052, but I see you've already used the fixed dates. Thanks.

PR #1054, opened 2024-08-05
Remove fitVids function call

GBKS:
This code calls a function that does not exist, resulting in a Javascript error on every page load.

[fitVids](http://fitvidsjs.com) is a jQuery plugin that probably was part of the codebase at some point (but definitely is not now).

I don't think there are any videos on the site. I also think that there are modern CSS-based approaches to resizing videos that could be implemented if necessary.


fanquake (approved):
ACK 1d5e611a04027228fb194a1d762f5028ed158967

PR #1073, opened 2024-10-07
RPC docs: fix broken links to tags >= v22.0

### LINKED ISSUE #1071: Links from the generated RPC docs back to version-tag are broken
Example:
https://bitcoincore.org/en/doc/28.0.0/ currently has a link in the left column:
> Documentation exported from [Bitcoin Core 28.0.0](https://github.com/bitcoin/bitcoin/tree/v28.0.0)

The URL is https://github.com/bitcoin/bitcoin/tree/v28.0.0 but the URL for the correct tag is https://github.com/bitcoin/bitcoin/tree/v28.0, so GitHub gives users 404s.

URLs for versions using the old 0.x.y scheme work.

stickies-v:
Since v22, the patch number is omitted in git tags, unless it is non-zero.
```
  version  -> tag
  0.21.0   -> v0.21.0
  22.0.0   -> v22.0
  24.0.1   -> v24.0.1
```

This breaks the "Documentation exported from Bitcoin Core 28.0.0" link on the RPC documentation page: https://bitcoincore.org/en/doc/28.0.0/ (and other >= v22 pages)

Fix this by automatically removing the patch if both the minor and patch numbers are zero. This creates some inconsistency in that on the same page we'll refer to both "28.0.0" and "28.0", but that seems unavoidable if we don't want to break permalinks.

Alternatives considered:
1. for affected version, update the `btcversion` field in existing RPC docs and in `generate.go`, but that would change a bunch of permalinks, making them not very perma
2. for affected versions, introduce a new `git_tag` field and update `generate.go` to produce that alongside `btcversion`, but it's a lot more involved than the automatic parsing in this PR
3. add the `x.x.0` tags in `bitcoin/bitcoin` and update the release process going forward, but I think there'd be hesitancy adding new tags just for this

Fixes #1071

### DIFF
--- a/_layouts/doc.html
+++ b/_layouts/doc.html
@@ -24,7 +24,14 @@
     <div>
       <p>
         {% if page.btcversion != "index" %}
-          Documentation exported from <a href="https://github.com/bitcoin/bitcoin/tree/v{{page.btcversion}}">Bitcoin Core {{page.btcversion}}</a>
+          <!-- Since v22, the patch number is omitted in git tags, unless it is non-zero:
+            version  -> tag
+            0.21.0   -> v0.21.0
+            22.0.0   -> v22.0
+            24.0.1   -> v24.0.1
+          -->
+          {% assign tag_version = page.btcversion | replace: ".0.0", ".0" %}
+          Documentation exported from <a href="https://github.com/bitcoin/bitcoin/tree/v{{ tag_version }}">Bitcoin Core {{ tag_version }}</a>
         {% endif %}
       </p>
     </div>


hodlinator:
Concept ACK

Thanks for fixing this! If anyone happens to have a nix-shell for building the website I'm happy to test, otherwise I am not sure when I would have time to fully verify it.

stickies-v:
> Did you have a look at other possible uses of `page.btcversion`?

Yeah, I outlined this in the PR description. It is used for permalinks, for example. That's why I'm not changing it in this approach. Or do you mean something else?

achow101:
ACK 24e7f4522494774ce841cb8d7178bd43e7117460

PR #1079, opened 2024-11-04
advisories: clarify low severity policy

fanquake:
Clarify that low severity issues will be disclosed after a fix exists in the current major release branch.

For example, if an issue is fixed in master, and then released in v29.0, if would be pre-announced, at the time of the v29.0 release, and made public 2 weeks later.

If an issue was fixed in master, it could also be backported to the current release branch, i.e 28.x, and then be in the next release of that branch, i.e v28.1. In that case, it would be pre-announced with the v28.1 release, and made public 2 weeks later.


dergoegge (approved):
ACK 34c56736d71623ad1cb3070f7adefc62e5632dd9

darosior (commented):
ACK 34c56736d71623ad1cb3070f7adefc62e5632dd9

PR #1124, opened 2025-04-17
Fix code block rendering in 29.0 Release Notes

hebasto:
On the master branch @ 56264987d13d18d6a80678cb82bf676fb8a61b30, the code block is incorrectly rendered as a single line:

image

This PR ensures it is displayed correctly over three lines:

image

Numbering has been replaced with bullets, as the rendering engine otherwise restarts numbering from 1 after the code block.

PR #1125, opened 2025-04-17
29.0: Fix code block in numbered list

murchandamus:
Fix the code block rendering to correctly show the three commands on separate lines.

Alternative to #1124


hebasto (commented):
Tested c38dfae9d967b9d72c573b31b9a9057252c29e7b. It doesn't seem to be working:

image

glozow:
Closing in favor of #1125

murchandamus:
That’s odd, to me [it looks](https://github.com/bitcoin-core/bitcoincore.org/blob/c38dfae9d967b9d72c573b31b9a9057252c29e7b/_releases/29.0.md) like this:
image

hebasto:
Could it depend on a browser? Mine is Firefox on Ubuntu 24.04 / 24.10.

murchandamus:
I’m on Ubuntu 24.04 with Firefox here as well. The preview looked quirky, but if I "View File", it looks correct as shown in the previous comment. Either way, if a bullet list is preferred, go with #1124 by all means. I just thought that this would be a smaller change that should also fix the main issue.

achow101:
The way the page is rendered in Github/your own markdown rendered doesn't matter, what matters is how it looks when the HTML is generated by the website builder.

murchandamus:
I see, thanks for the explanation, @achow101.

PR #1133, opened 2025-05-17
blog: add redirect for disclose-cve-2024-52919 post

fanquake:
I changed the URL of this post, but that was after it was posted to the mailing list, this has caused some amount of confusion. Add a redirect, so if anyone hits the original link, they'll end up in the right place.

See: https://groups.google.com/g/bitcoindev/c/v2mwcWtgfxM.


glozow (commented):
ACK 48112011f09d390cd3c5cba18818f55103892600

tested locally that /en/2025/03/31/disclose-cve-2024-52919 redirects

PR #1155, opened 2025-06-29
Fix broken Transifex link

azuchi:
I fixed broken Transifex link based on the fix in #1135.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -2,7 +2,7 @@
 
 title: Bitcoin Core
 url: https://bitcoincore.org
-transifex_url: https://www.transifex.com/bitcoincore/bitcoincoreorg
+transifex_url: https://explore.transifex.com/bitcoin/bitcoin/
 
 # Jekyll configuration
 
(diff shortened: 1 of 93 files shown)


_config.yml
```diff
@@ -2,7 +2,7 @@
 
 title: Bitcoin Core
 url: https://bitcoincore.org
-transifex_url: https://www.transifex.com/bitcoincore/bitcoincoreorg
```
murchandamus, line 5: I was wondering whether this was replaced mistakenly, but it is also 404, and while there is a `https://explore.transifex.com/bitcoincore/` org on Transifex, I could not find a "bitcoincoreorg" project in it.

murchandamus (commented):
Tested that all three replaced links 

- https://www.transifex.com/projects/p/bitcoin
- https://www.transifex.com/bitcoincore/bitcoincoreorg
- https://www.transifex.com/bitcoin/bitcoin/

are 404 and the replacement works and appears to be pointing at the Bitcoin Core translation project.

I’m still wondering whether there is a separate translation project for bitcoincore.org, but if so, I could not find it. It seems that it used to exist given the prior mention in #670.

hodlinator:
Close this since #1231 was merged?

Issue #1162, opened 2025-08-12
Including the built HTML in repo


achow101:
One of the issues we have had with the website build server is that it occasionally fails to build the site, and resolving those issues is actually quite annoying. One suggestion that @TheBlueMatt has is that we should be committing the built HTML into the repo, and deployment for the website can be reduced to running `verify-commits.py` and then copying the HTML.

This does give us several advantages:
* since we won't be limited by what dependencies can be installed on the build machine, we can use whatever build dependencies we want
* reduces the attack surface on the website infrastructure
* makes maintaining the website infrastructure easier

The main drawback with this process would be requiring all contributors to commit the built HTML in their PRs. Given that we do get contributors who are not necessarily developers so may not have the build environment setup, this could be an additional burden that reduces contributions of things like translations. This could be mitigated by having a CI task or bot which does the build and opens a PR containing the HTML after something is merged. It may be possible to also configure the maintainer merge script to perform and commit the build during the merge, but I'm not sure if that is desirable.

Any opinions?

ryanofsky:
> The main drawback with this process would be requiring all contributors to commit the built HTML in their PRs. 

Would another drawback be that this would require people to manually review the generated code?

achow101:
> Would another drawback be that this would require people to manually review the generated code?

I don't think so. It should be possible to reproducibly build the site so review can be building the site and checking the diff matches. We should be able to do that as part of CI.

achow101:
> I'm not sure why getting a CI job that generates the site to work reliably would be easier than getting the "website build server" to work reliably, but I could imagine that being the case.

The build server needs to run in close proximity to the actual web server. Because of that, it has heightened security requirements which limits where we can install dependencies from, and what those are. The result is really that we can't/don't use bundler and ruby gems to install the dependencies which makes building the site much more annoying.

In contrast, CI can use bundler and ruby gems to do the build since the hardware is neither ours nor has those security requirements.

achow101:
Another downside is that all PRs will have a merge conflict with each other on the html, so I think having everyone include the HTML might not be viable.

PR #1163, opened 2025-08-13
Commit built html

achow101:
Implements #1162

It turns out the build was not entirely deterministic as the build time and timezone of the building machine would be included in several places. The first several commits makes it so that those timestamps use the commit timestamp.

After including all of the generated HTML, the CI is changed to check that it replicates the HTML that is in the repo.

Lastly, the build has had a deprecation warning which is resolved in the last commit. No HTML changes to fix that warning.


achow101:
> What would we need to do differently before we open a PR?

Just building the site and committing the result. Can be done easily with `make`.

achow101:
> Am I understanding correctly that if I checkout this PR, and run `make build`, there should be no diff generated? As that currently isn't the case.

It's supposed to. What does your diff look like?

PR #1164, opened 2025-08-23
Use a layout for releases that includes the publication date

### LINKED ISSUE #1159: Release notes should display release date
The release notes pages are great (eg [28.2](https://bitcoincore.org/en/releases/28.2/) except that they don't include the release date. Some older pages on bitcoin.org do, eg [0.21.0](https://bitcoin.org/en/release/v0.21.0), for comparison, though not recent ones. The [markdown files](https://github.com/bitcoin-core/bitcoincore.org/blob/master/_releases/28.2.md) do include a date, so I guess this just isn't being exposed via the layout logic that marks up releases as pages:

https://github.com/bitcoin-core/bitcoincore.org/blob/07ca5bf4db60958b0f7e89b160552a6754ca2daa/_config.yml#L51-L57

The post.html layout exposes date information I think:
https://github.com/bitcoin-core/bitcoincore.org/blob/07ca5bf4db60958b0f7e89b160552a6754ca2daa/_layouts/post.html#L52-L64

Perhaps something similar could just be added to page.html? Or, maybe a custom layout for releases would make sense?

achow101:
The publication date is already included in the `post` layout. This PR extracts that into an include to be used in a custom `release` layout that inherits the `page` layout that releases currently use.

Closes #1159

### DIFF
--- a/_posts/es/releases/2021-09-29-release-0.21.2.md
+++ b/_posts/es/releases/2021-09-29-release-0.21.2.md
@@ -6,7 +6,7 @@ permalink: /es/releases/0.21.2/
 excerpt: Bitcoin Core versión 0.21.2 ya está disponible
 date: 2021-09-29
 type: releases
-layout: page
+layout: release
 lang: es
 
 ## Use a YAML array for the version number to allow other parts of the
--- a/_posts/es/releases/2025-06-26-release-28.2.md
+++ b/_posts/es/releases/2025-06-26-release-28.2.md
@@ -6,7 +6,7 @@ permalink: /es/releases/28.2/
 excerpt: Bitcoin Core version 28.2 ya está disponible 
 date: 2025-06-26
 type: releases
-layout: page
+layout: release
 lang: es
 
 ## Use a YAML array for the version number to allow other parts of the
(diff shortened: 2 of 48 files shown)


murchandamus:
ACK a5c1f956db43d428db524a68b0f75e4fda9e16a5

Pretty, pretty please, yes! I have gone to the blog to find the date of a release after looking at the release notes more often than I care to admit.

achow101:
Rebased and applied the scripted-diff to the recent releases.

> It only produces a little message like "Bitcoin Core 30.2 was published on January 10, 2026 ." at the bottom of the file, which maybe would be nicer as "Bitcoin Core 30.2 was released on January 10, 2026" in normal text at the top of the file.

This footer is shared with non-release posts. Additionally, putting it at the top is much more involved since the header of the post comes from the release's .md file so it's way more annoying to do it in a way that doesn't require the release post author to remember to add specific lines.

> The space before the full-stop can be avoided by writing:

Done

ajtowns:
ACK 633ba81a8ee9f1d6d5ad6fa050065e1b060027e3

> Additionally, putting it at the top is much more involved since the header of the post comes from the release's .md file so it's way more annoying to do it in a way that doesn't require the release post author to remember to add specific lines.

I asked claude code, and it suggests:

```diff
--- a/_layouts/release.html
+++ b/_layouts/release.html
@@ -2,6 +2,8 @@
 layout: page
 ---
 
+<p><b>Released: <time class="dt-published" datetime="{{ page.date | date_to_xmlschema }}">{{ page.date | date: "%B %d, %Y" }}</time></b></p>
+
 {{ content }}
 
 {% include posted-date.html %}
```

which seems to do the right thing to me. Having the date on the page at all is a bigger deal than where/how it appears though.

PR #1166, opened 2025-08-23
Remove unused files

achow101:
These files appear to be unused and removing them doesn't make any difference to the built site.


stickies-v (approved):
tACK 00c0857485e62e43e04c3168f5a1fe9a82d5ef3c

PR #1173, opened 2025-09-22
fix broken mobile layout on contact page

ritorhymes:
## Description

A table and an inline code span are breaking the layout on the contact page at the mobile breakpoint. They are both overflowing past the right page margin and causing the entire page to increase its width and become scrollable; this breaks the header and footer by creating a blank white gutter on the right margin.

**This fix:**
- Make tables horizontally scrollable to contain overflow content.
- Forces the specific inline code span to wrap down to the next line.

The fix for each element is added to:
`site.scss` for the table fix in the section for styling tables
`typography.scss` for the inline code span fix in its code block styling section

## Before and After

**Before (Notice the footer broken because the page width expands past the normal margin)**
contact-before

**After**
contact-after-prescroll

**After - Scrolling inside the table to show content contained**
contact-after-postscroll


ritorhymes:
So as far as I've seen, the download page has the same page margin issues on mobile with the inline code element. The very first inline code about gpg would need to be replaced. And when you click to reveal the content of the OS 'verification instructions', each one has a really long fingerprint inline code which would need to be replaced too.

After making those changes to that page, I think that would be it. I can then officially certify its mobile responsiveness 🧾 and then be on my way.

I think the change to the fingerprint's characters for spacing is probably fine.. but at the same time I feel like someone out there might treat it like nuclear launch codes and erupt...lol

PR #1178, opened 2025-10-06
Reduce the prominence of examples in the security advisories page

darosior:
In #1147 we added examples for each severity level. While this is useful to have as an illustration in the policy, this made the first part of the "Policy" section take a lot more space. It takes almost 1.5x the height of my high resolution office monitor. This means scrolling up more than one entire page before getting to the disclosure timeline, let alone the latest publish advisories.

This PR reduces the prominence of the severity level by reducing the size of the section titles and by putting the example under a collapsible section. 

Here is how it changes the display for me.

Before:
image

After:
image


darosior:
I feel like we could almost get back to bullet points for each severity level. It's a bit contrived, but i feel like the flow makes more sense this way. Let me know what you think.

image
image

janb84:
I would opt for the bullet points style for increased readability. less scrolling and better overview over the different levels.

janb84 (commented):
ACK 0e8f0c84b4673b0d23a7b810bd9c17efdeff4318

PR introduces bullet points for the severity headers and collapsible section for the examples. This makes the page more readable and creates a more unified severity section, with the different levels displayed together.

- code review ✅

PR #1180, opened 2025-10-07
navigation: move "security advisories" to a top-level entry

darosior:
For a security-critical project like Bitcoin Core, it makes sense to have the security advisories page be a main entry in the navigation menu. Furthermore, the development dropdown was not the most appropriate place for the security advisories entry and made it harder to find for users.

Here is a screenshot of the result:
image


janb84 (commented):
NIT: Also move the menu item for the non english languages, e.g. line 82

janb84 (commented):
ACK 4f4fce54beff05e153d103d279d92ead0980159e

PR changes top menu, moves security-advisories from submenu to top menu. Given the security sensitivity of the project I agree that this should have a more prominent place.

PR #1184, opened 2025-10-10
contrib: add a Dockerfile to run the website locally

darosior:
It was useful to me, so i cleaned it up and PR'd it in case others do not have a full Ruby toolchain installed, but do have Docker.

### DIFF
--- a/contrib/devtools/Dockerfile
+++ b/contrib/devtools/Dockerfile
@@ -1,0 +1,17 @@
+FROM ruby:3.1.2
+
+ENV NOKOGIRI_USE_SYSTEM_LIBRARIES=true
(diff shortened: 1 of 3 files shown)


contrib/devtools/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ruby:3.1.2
+
+ENV NOKOGIRI_USE_SYSTEM_LIBRARIES=true
+ENV JEKYLL_ENV=production
+ENV RUBYOPT="-KU -E utf-8:utf-8"
+
+RUN apt update && apt install -y make
```
stickies-v: We don't use make at all, so this line be removed entirely

PR #1188, opened 2025-10-14
fix: Correct lifecycle dates for 31.x

murchandamus:
Fixes the off-by-one in the maintenance lifecycle for the 31.x major branch.


achow101:
ACK 11d6079be9e8adf181aff00516a76d2551b88154

PR #1200, opened 2025-11-24
Update the release lifecycle page to reflect current practices

darosior:
This updates the release lifecycle page to more closely reflect our actual releases. This started as an effort to drop the superfluous EOM status, but ended up as a rewrite of the Versioning, Maintenance and Schedule sections since all were depending on each other.

We only make a distinction between maintained and EOL versions. The EOM status adds unnecessary complexity and is misleading, so remove it.

This also cleans up the Versioning section, which was using three major heading (Versioning, Major releases and Maintenance releases) for a small amount of content all on the same topic of how the software we release is versioned. This also cleans up a redundant statement about backporting consensus changes and removes an unnecessary statement about the distinction between minor and major features for backports (which we don't even follow nowadays as far as i know).

Then we rewrite the maintenance version, which was making a distinction between "maintenance end" and "end of life" that we do not follow in practice, and overall makes it clearer and more concise. In this section i also remove some confused language about how we handle security fixes, which is not consistent with our practice, in favour of a general sentence that EOL versions do not receive security fixes and a link to our Security Advisories page which gives more details on this topic.

Wherever i touched i also updated the examples to use newer versions. In some cases i also switched from pointing to the exact major release (X.0) to the major version (X.Y) which is more accurate and also what we already do on the security advisories page.


darosior:
Here is how it looks like after this change
image

darosior:
Some archeology:
- The EOM status was introduced in https://github.com/bitcoin-core/bitcoincore.org/pull/37 (09593208)
- The original EOL policy was discussed in [this IRC meeting](https://www.erisian.com.au/bitcoin-dev/log-2016-01-21.html#l-1185), though as far as i can tell there was no mention of the EOM status. My understanding was that participants essentially discussed what is in effect our currently observed EOL policy but only for the past two major versions (what's as EOM on the website)

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -16,44 +16,33 @@ This document describes the life-cycle of the Bitcoin Core software package rele
...
-We aim to make a major release every 6-7 months.
-
-These will be numbered 22.0, 23.0 etc.
-
-## Maintenance releases
-
-We will provide maintenance "minor releases" that fix bugs within the major releases. As a general rule we do not introduce major new features in a maintenance release (except for consensus rules). However, we may add minor features where necessary, and we will back-port consensus rule changes such as soft forks.
-
-Minor releases will be numbered 22.1, 22.2, 23.1, 23.2 etc.
+We will provide maintenance "minor releases" that fix bugs (security and otherwise) within the major releases. These
```
fanquake:
```suggestion
We will provide maintenance releases, that fix bugs (security and otherwise), for each major release. These
```

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -16,44 +16,33 @@ This document describes the life-cycle of the Bitcoin Core software package rele
...
+"End of Life". The threshold for backporting a change to a past major version increases as it ages. For example, if the
+last major release is 30.0, then 29.x and 28.x are also considered maintained. Once 31.0 is released, then 28.x becomes
+"End of Life".
 
-For example, major version 22.0 was released on 2021-09-13 and we provided maintenance fixes (point releases) until 2022-12-14.
-Critical security issues would still be continued to be fixed until the EOL date of 2023-04-01.
-However, to take advantage of bug fixes, you would have to upgrade to a later major version.
+Major versions that are "End of Life" do not, as a general rule, receive security fixes. For more about our policy on
+security fixes, see our [security advisories](/en/security-advisories/) page. We recommend running the latest
+maintenance release (point release) of the most recent major version you are able to upgrade to.
```
murchandamus: Previously, this is referred to as “minor release”. I would recommend using the same term to refer to the same concept:

```suggestion
maintenance release (minor release) of the most recent major version you are able to upgrade to.
```

murchandamus (commented):
ACK 70e2425ffefea08467d1985409a6c33a3143e850,

with a few optional suggestions.

_posts/en/pages/2016-01-15-lifecycle.md
```diff
@@ -16,44 +16,33 @@ This document describes the life-cycle of the Bitcoin Core software package rele
...
-Please note that minor versions get bugfixes, translation updates, and soft forks. Translation on [Transifex][bitcoin-transifex-link] is only open for the last two major releases.
+We always maintain the past three major versions released. Once a major version falls outside that window, it becomes
+"End of Life". The threshold for backporting a change to a past major version increases as it ages. For example, if the
+last major release is 30.0, then 29.x and 28.x are also considered maintained. Once 31.0 is released, then 28.x becomes
+"End of Life".
 
-For example, major version 22.0 was released on 2021-09-13 and we provided maintenance fixes (point releases) until 2022-12-14.
-Critical security issues would still be continued to be fixed until the EOL date of 2023-04-01.
-However, to take advantage of bug fixes, you would have to upgrade to a later major version.
+Major versions that are "End of Life" do not, as a general rule, receive security fixes. For more about our policy on
```
murchandamus: I don’t think it’s necessary to point out that we might make an exception and do more work than we generally intend to:

```suggestion
Major versions that are "End of Life" do not receive security fixes. For more about our policy on
```

willcl-ark:
Concept ACK, seems like a nice improvement.

I'd be happy to see this go in as-is after the above comments have been addressed, but I also took a stab at improving the general readability of the page a bit myself in https://github.com/willcl-ark/bitcoincore.org/commit/58bbe58582879d858e8a735db3e5e7186991d782, or in rendered form at https://github.com/willcl-ark/bitcoincore.org/blob/pr-1200/_posts/en/pages/2016-01-15-lifecycle.md, as I was reviewing the overall document.

Feel free to take any changes from there that you like.

_posts/en/pages/2016-01-15-lifecycle.md
darosior: I think your alternative suggests that End of Life versions would never receive a security fix, no matter how critical. I think that's incorrect and my sentence better reflects our practices without setting too much expectations, so i didn't take your suggestion.

murchandamus: Perhaps:

```suggestion
Major versions that are "End of Life" generally do not receive security fixes. For more about our policy on
```

PR #1221, opened 2026-02-10
Add release artifacts documentation page

PauloAAlmeida:
Add a page explaining the different file types distributed on the Bitcoin Core download page and in release torrents.

Closes #1220

### DIFF
--- a/_posts/en/pages/2026-02-10-release-artifacts.md
+++ b/_posts/en/pages/2026-02-10-release-artifacts.md
@@ -12,7 +12,7 @@ Bitcoin Core release files are hosted at [https://bitcoincore.org/bin/](https://
 
 ## Build Suffixes
 
-- **Standard Binaries** (`.tar.gz`, `.zip`, `-setup.exe`): Production-ready, signed binaries for general use.
+- **Standard Binaries** (`.tar.gz`, `.zip`, `-setup.exe`): Production-ready binaries for general use.
 - **`-debug`**: Includes symbols for backtrace generation. Larger file size and reduced performance.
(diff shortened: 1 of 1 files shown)


_posts/en/pages/2026-02-10-release-artifacts.md
```diff
@@ -0,0 +1,29 @@
...
+permalink: /en/release-artifacts/
+version: 1
+---
+
+Bitcoin Core release files are hosted at [https://bitcoincore.org/bin/](https://bitcoincore.org/bin/) and distributed via torrent.
+
+## Build Suffixes
+
+- **Standard Binaries** (`.tar.gz`, `.zip`, `-setup.exe`): Production-ready binaries for general use.
+- **`-debug`**: Includes symbols for backtrace generation. Larger file size and reduced performance.
```
willcl-ark: Sorry, but I still don't think this is exactly correct.

In the guix builds we produce a single set of bins, using `-O2 -g`, then post-process them using `split-debug.sh`. This splits off the symbols into a standalone dbg file. The bins are then stripped, and two archives are created, one with the binaries, and one with the debug symbols.

I don't think it's accurate ∴ to say that the debug bins have "reduced performance", as they're the same bins; the debug archive only contains the symbols IIUC.

cc @fanquake

PauloAAlmeida: @willcl-ark please let me know if you need any adjustments. Thank you for your review.

achow101:
triggering ci

PauloAAlmeida:
> triggering ci

Thanks. CI failures appear to be unrelated to this PR. Both failing jobs hit the same pre-existing issue.

PR #1223, opened 2026-02-10
test: Ignore $(package) when checking for liquid templates

achow101:
$(package) is not used in any liquid template strings, but does appear in the title of a PR. Ignore this string so that CI does not fail on a false positive on these strings.

Fixes the current CI failures.


fanquake:
Is this still needed?

achow101:
Yes... Some time between this PR being opened and now, `make test-fast` stopped emitting an error when it should have.

achow101:
The issue was with how xargs exits when it has to break the command into multiple commands. I've pushed a commit which will fix this issue.

Sjors:
Seems like a good idea, but the finer nuances of xargs and grep are a bit over my head.

fanquake:
ACK 0e160fa56bc1597af416bb2796bdbcdd71ebde4e

PR #1231, opened 2026-03-23
doc: fix broken Transifex URLs across release notes and config

Bortlesboat:
All `transifex.com` links in release notes and `_config.yml` currently return 404. Transifex migrated to `explore.transifex.com`.

This PR updates all affected files:
- `www.transifex.com/projects/p/bitcoin/` → `explore.transifex.com/bitcoin/bitcoin/`
- `www.transifex.com/bitcoin/bitcoin/` → `explore.transifex.com/bitcoin/bitcoin/`
- `www.transifex.com/bitcoincore/bitcoincoreorg` → `explore.transifex.com/bitcoin/bitcoin/`

Rebased version of #1155, which has had merge conflicts since June 2025. Credit to azuchi for identifying the issue.

Verified: old URLs return HTTP 404, new URL returns HTTP 200.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -2,7 +2,7 @@
 
 title: Bitcoin Core
 url: https://bitcoincore.org
-transifex_url: https://www.transifex.com/bitcoincore/bitcoincoreorg
+transifex_url: https://explore.transifex.com/bitcoin/bitcoin/
 
 # Jekyll configuration
 
--- a/_config.yml
+++ b/_config.yml
@@ -2,7 +2,6 @@
 
 title: Bitcoin Core
 url: https://bitcoincore.org
-transifex_url: https://explore.transifex.com/bitcoin/bitcoin/
 
 # Jekyll configuration
 
--- a/_posts/es/releases/2021-09-29-release-0.21.2.md
+++ b/_posts/es/releases/2021-09-29-release-0.21.2.md
@@ -141,5 +141,5 @@ Gracias a todos los que han contribuído directamente a esta edición:
 - W. J. van der Laan
 
 Así como a todos aquellos que han ayudado con las traducciones en
-[Transifex](https://www.transifex.com/bitcoin/bitcoin/).
+[Transifex](https://explore.transifex.com/bitcoin/bitcoin/).
 {% endgithubify %}
--- a/_posts/ja/releases/2017-09-01-release-0.15.0.md
+++ b/_posts/ja/releases/2017-09-01-release-0.15.0.md
@@ -867,6 +867,6 @@ Coin Age Priorityの削除 {#removal-of-coin-age-priority}
 - Warren Togami
 - Wladimir J. van der Laan
 
-[Transifex](https://www.transifex.com/projects/p/bitcoin/)での翻訳を手伝ってくれたみなさんもありがとうございます。
+[Transifex](https://explore.transifex.com/bitcoin/bitcoin/)での翻訳を手伝ってくれたみなさんもありがとうございます。
 
 {% endgithubify %}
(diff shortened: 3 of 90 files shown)


hodlinator (approved):
ACK b4e8e12fb3b9394a3742620db2a4e0a5c0dba896

Same file changes as in https://github.com/bitcoin-core/bitcoincore.org/pull/1228#pullrequestreview-3976406939.

_config.yml
```diff
@@ -2,7 +2,7 @@
 
 title: Bitcoin Core
 url: https://bitcoincore.org
-transifex_url: https://www.transifex.com/bitcoincore/bitcoincoreorg
+transifex_url: https://explore.transifex.com/bitcoin/bitcoin/
```
murchandamus: My review stumbled over this replacement [again](https://github.com/bitcoin-core/bitcoincore.org/pull/1155#pullrequestreview-2972236646). The original link is 404. I can find a ["bitcoincore" organisation](https://explore.transifex.com/bitcoincore/) on Transifex, but it has no projects in it. It looks like the previously linked project [was deleted in 2019](https://github.com/bitcoin-core/bitcoincore.org/issues/670).

It seems to me that the intent for this variable was to link to the __website translation project__ rather than the client software translation project, but the website [doesn’t use Transifex](https://github.com/bitcoin/bitcoin/issues/16754#issuecomment-526373630) (and never has?). So, I suspect that this link has always been broken.

From a quick grep in the website code, it appears that the variable `transifex_url` is only used in `_includes/istranslated.html` for a "not translated" macro.

Assuming my read is right, this variable could be renamed to `translation_url` (here __and__ in `_includes/istranslated.html`) and should link instead to the translation instructions for the website:

```suggestion
translation_url: https://github.com/bitcoin-core/bitcoincore.org/blob/master/CONTRIBUTING.md#translation-process
```

murchandamus (commented):
The replacements of 
- https://www.transifex.com/bitcoin/bitcoin/
↦ https://explore.transifex.com/bitcoin/bitcoin/
- https://explore.transifex.com/projects/p/bitcoin/
↦ https://explore.transifex.com/bitcoin/bitcoin/

Look good to me.

I think the `transifex_url` variable: https://www.transifex.com/bitcoincore/bitcoincoreorg should point at the website translation instructions instead. It should perhaps also be renamed as the website doesn’t use transifex.

_config.yml
Bortlesboat: You're right, this was for the website project, not Bitcoin Core's. I grepped the repo and `transifex_url` is only defined in `_config.yml` — it's not referenced anywhere else. `istranslated.html` doesn't exist in `_includes/` either. Looks like dead code.

Happy to rename it to `translation_url` pointing at the CONTRIBUTING.md translation process like you suggested, or just remove the variable entirely. What do you think?

murchandamus: Ah right, I had an old branch, and that file was removed in e327ae7282c410f98f14ba728a05b66d4d8236a4 per #1166. I assume the config line can then also be removed, but @achow101 would know better.

murchandamus:
```suggestion
```

Bortlesboat: Done, removed the line entirely. Confirmed there are no remaining references to `transifex_url` anywhere in the repo.

hodlinator (approved):
re-ACK aa807f303ffebcc22d22cc0c743e4abb800213e7

`transifex_url` was added in 2bf1bc7791c1e2e96774479c11b0e74c7b1ac48c and used in _includes/[_]istranslated.html which was later removed in e327ae7282c410f98f14ba728a05b66d4d8236a4.

If you happen to re-push, I would switch the order of the commits so you remove `transifex_url` without first correcting it.

murchandamus (commented):
I agree that the reversing the order of the two commits would be preferable, but the website code doesn’t get that enough traffic that I’d consider it important.

ACK aa807f303ffebcc22d22cc0c743e4abb800213e7

_config.yml
achow101: This line can be removed.

achow101:
ACK aa807f303ffebcc22d22cc0c743e4abb800213e7

PR #1236, opened 2026-04-09
ci: Don't use a container for verify-commits task

achow101:
Apparently using a container with a specific image causes the `.git` directory to disappear. `verify-commits.sh` can run on ubuntu so the container can be dropped to resolve that issue. Also switching to ubuntu requires `sudo`ing `apt-get`.

A run of this task is available at https://github.com/achow101/bitcoincore.org/actions/runs/24218968381/job/70705928687, although that has an additional change to have it run on a non-master branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,8 +52,6 @@ jobs:
   verify-commits:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
-    container:
-      image: debian:latest
     steps:
       - uses: actions/checkout@v6
         with:
@@ -61,8 +59,8 @@ jobs:
 
       - name: Install packages
         run: |
-          apt-get update
-          apt-get install -y \
+          sudo apt-get update
+          sudo apt-get install -y \
             git \
             gnupg
 


.github/workflows/ci.yml
```diff
@@ -52,17 +52,15 @@ jobs:
   verify-commits:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
-    container:
-      image: debian:latest
```
maflcko, line 56: Not sure if this is important, but I presume it could make sense to use the exact same debian version that the infra is running on, to be more precise in catching debian-version-specific issues. Maybe `debian:bookworm`, like above?

But I guess it would be rare for such an issue to exist, so no strong opinion, just a nit.

.github/workflows/ci.yml
```diff
@@ -52,17 +52,15 @@ jobs:
...
     steps:
       - uses: actions/checkout@v6
```
maflcko, line 56: I think an alternative fix would be to just move the checkout step to after the install step. IIRC it will see `git` installed and then use that.

maflcko (approved):
lgtm, but I left two nits. Feel free to ignore.

.github/workflows/ci.yml
achow101: It isn't that `git` is not installed, but rather that git is not recognizing the working directory as being a git repo, as the `.git` directory is missing.

.github/workflows/ci.yml
achow101: It doesn't matter for verify commits.

For some reason, not using a container fixes the particular CI error, see https://github.com/actions/checkout/issues/363

.github/workflows/ci.yml
maflcko: Are you sure. The log says that git is not installed:

https://github.com/bitcoin-core/bitcoincore.org/actions/runs/24216061534/job/70696772544#step:3:27

```
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Downloading the archive
Writing archive to disk
Extracting the archive
/usr/bin/tar xz --warning=no-unknown-keyword --overwrite -C /__w/bitcoincore.org/bitcoincore.org/7af2c3ca-f186-44ac-a8f2-22a29ee53986 -f /__w/bitcoincore.org/bitcoincore.org/7af2c3ca-f186-44ac-a8f2-22a29ee53986.tar.gz
Resolved version bitcoin-core-bitcoincore.org-1599ca8

achow101: Ah, so it is being checked out without git, hence why the script later fails saying that `.git` doesn't exist.

Either way, I don't think it matters whether this task is run inside a container or not. The script is distro independent and doesn't rely on specific package versions like the website build does.

maflcko: Right, I've already approved it, so you can merge it. I just wanted to fully explain the issue :sweat_smile:

maflcko:
lgtm ACK aa4cf6848a2853318c6c67c64f9e2bb98c0800da fwiw

PR #1239, opened 2026-04-21
maintenance-table: update for 31.x release

fanquake:
Now that `31.0` is released:
* Add placeholder for 32.x
* Add `28.x` EOL date

PR #1241, opened 2026-04-21
ci: check maintenance table updated

stickies-v:
A very rudimentary check to verify that the maintenance table is updated when a new major release is done (see e.g. #1239 ). Does not validate any details to keep the script simple, just that the latest major release has a date associated with it.

The minimal checks are intentional, to avoid CI getting in the way when the release process is a bit non-standard.

Can be tested locally by modifying the `maintenance-table.md` and running `./contrib/check-maintenance-table.sh`.


achow101:
ACK 8d196d3e19b4ab51e1aff34b5e385b67eedc6df8

PR #1246, opened 2026-05-11
Bump Ruby to 3.3.8 and Debian to Trixie

Sjors:
The current Ruby version is end-of-life and doesn't work on modern macOS, making it hard to test changes locally. The latter can be worked around, see https://github.com/bitcoin-core/bitcoincore.org/pull/1246#issuecomment-4421636547.

Debian Bookworm is still in LTS, but so is the newer Trixie. Hopefully the backend can be updated to that.

The latest patch version is 3.3.11, but a backend upgrade to Trixie ships Ruby 3.3.8. The 3.3.x branch is [still maintained](https://www.ruby-lang.org/en/downloads/branches/), until March 2027.

I lightly tested the site locally.


fanquake:
> and doesn't work on modern macOS, making it hard to test changes locally.

You can test changes using https://github.com/bitcoin-core/bitcoincore.org/blob/master/contrib/devtools/README.md#dockerfile, which should work fine on macOS.

Sjors:
The sass warning reveals a bigger can of worms: we're relying on LibSass which has been [abandoned in favor of Dart Sass](https://sass-lang.com/libsass/).

So I dropped 34afa34823682369617a6e25246fad67ad877156. The warnings aren't too bad because they mostly get suppressed:

> Warning: 23 repetitive deprecation warnings omitted.

A real fix would involve installing Dart Sass on CI, brining the commit back and then decide whether we want to make the same (trivial) changes to the vendored code.

Sjors:
These sass warnings predate this PR, so I'll track them separately: #1249

Sjors:
Just saw #1223, which is probably better than my second commit. Marking this draft.

Sjors:
Temporarily pushed the commits from #1223 here to see if those still work after the Ruby / Debian update.

achow101:
The debian only CI job is setup to match the current build infrastructure to let us know if a change may break when the website is rebuilt. It should only be changed when the infra is changed.

Sjors:
@achow101 I know that, this PR implies changing the infrastructure from Debian to Trixie. If CI passes, that change should be safe. Well, at least not obviously broken.

Sjors:
Rebased after https://github.com/bitcoin-core/bitcoincore.org/pull/1223 landed and slightly clarified the PR description.

PR #1253, opened 2026-05-15
sass: Resolve deprecated slash as division

achow101:
Resolves the deprecation warning that building emits.


achow101:
Huh, can't do this while we still require debian-only building to work.

PR #1267, opened 2026-07-14
Render structured RPC docs from OpenRPC specs

satsfy:
Forward note: the diff is big but most of it is auto-generated json and html files, pls read commit by commit to review.

The RPC reference pages are a plaintext dump of `bitcoin-cli help` today. This reworks contrib/doc-gen to read committed OpenRPC specs and render structured pages with an arguments table and a typed result tree. Every element in the tree, aside from root, is collapsible.

----

image

image

image

----

A recent Open PR in [Bitcoin Core](https://github.com/bitcoin/bitcoin/pull/34683) introduces OpenRPC spec schema for all RPCs of Bitcoin Core using core's own documentation. I have backported it to applicable version of Bitcoin Core (v31 - v26) and generated OpenRPC schemas with `bitcoin-cli getopenrpcinfo` on each, producing JSON files. Further backports from v25 required workarounds that caused the output to become progressively more unsafe, so earlier versions keep only the current plaintext rendering.

The old plain-text format from help stays available in the UI, just one click away, as well as the OpenRPC spec. Each RPC method URI has a companion plain-text page at `.../<name>/raw/` and `.../<name>/openrpc/`, reached from a toggle next to the title. The switch UI element is reactive to screen size, assuming a separate row when less space is available.

The field "examples" is not exported in openrpc, so I take them from the existing plaintext spec. The same can be said for `named-arguments` annotations.

Additionally, I added a copy button to the OpenRPC and Plain Text textboxes. Just felt like it should be there.

----

image

image

-----

Note on maintenance: while `getopenrpcinfo` RPC is not merged in Bitcoin Core, the process of generating a new spec will have to be manual (and I volunteer to do it in the meantime). When it is available, I can leverage the documentation directly using the `contrib/doc-gen/generate.go` node on regtest.

This PR was created using Claude Opus 4.8 and reviewed by me.

It's my first time contributing here so, if I'm missing something important, please let me know.

EDIT: all the pictures I had taken on this PR intially was using a dark mode extension, so they were dark. I didn't actually implement dark mode. Edit updates the pictures.


satsfy:
Sorry to bother, @fanquake, but I see you are the person merging most PRs in this repo. This PR is much longer and harder to review than the usual ones here, also modifies the essence of the RPC doc page.

If at all possible, I'm seeking a concept ACK.

How can I make the review here easier?

satsfy:
Rebase reordered commits such that html generation comes last.

maflcko:
Not sure why this needs to be rushed, especially with backports to older releases. This is basically impossible to review.

Also, there are still many bugs/shortcomings, like https://github.com/bitcoin/bitcoin/pull/35836 that will all need to be fixed first.

willcl-ark:
Would it be more sensible just to use this for 32.x onwards? I am asking as I'm pretty ignorant of how the website works, but ISTM that this would likely make it a smaller/more reviewable change?

achow101:
> Would it be more sensible just to use this for 32.x onwards?

Would prefer that.

satsfy:
Sounds great, will do. I think we are missing the `getopenrpcinfo` new RPC on [v32 release notes](https://github.com/bitcoin-core/bitcoin-devwiki/wiki/32.0-Release-Notes-Draft).

willcl-ark:
Thanks @satsfy added a note to https://github.com/bitcoin-core/bitcoin-devwiki/wiki/32.0-Release-Notes-Draft#new-rpcs

PR #1275, opened 2026-08-27
Add the missing default social preview image

devdavidejesus:
The og:image / twitter:image fallback points at assets/images/default-thumb.png, which does not exist, so bitcoincore.org links shared on social media render with no preview image.

- Adds the missing file under the exact name the template already references: a 1200x630 card built from the project's own marks (the ring icon and the Bitcoin Core wordmark), the About page's first sentence, and the genesis block hash.
- Switches the fallback twitter:card to summary_large_image so the full-width card is used — matching what the template already does for pages with a feature image.


devdavidejesus:
How the card renders (1200x630): 
bitcoincore-social-card

devdavidejesus:
What do you think of this as the default social preview image?

I tried to keep the design minimal and consistent with the Bitcoin Core branding, while giving bitcoincore.org links a proper preview when shared on social media.

@murchandamus @brunoerg — any feedback is welcome.

PR #1277, opened 2026-09-03
download: Pass checksum file explicitly to GPG

jonasnick:
The gpg man page strongly discourages relying on GPG's historical behavior of finding a detached signature's data file by stripping the signature filename's suffix.

Pass SHA256SUMS explicitly when verifying SHA256SUMS.asc, as recommended by the man page.

Only tested on linux with
```
❯ gpg --version
gpg (GnuPG) 2.4.9
```
but it should apply to distributions of the GnuPG gpg command on all supported operating systems. https://www.gnupg.org/documentation/manuals/gnupg/Operational-GPG-Commands.html

### DIFF
--- a/_includes/templates/download.html
+++ b/_includes/templates/download.html
@@ -165,7 +165,7 @@
 
       <li>{{page.verify_checksums_file}}
 
-        <pre class="highlight"><code>{{GPG}} --verify SHA256SUMS.asc</code></pre></li>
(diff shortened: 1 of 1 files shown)


murchandamus (commented):
Concept ACK

fanquake:
ACK 11ce6549ffb960ca9e829a1125245788e78c1294

PR #1278, opened 2026-09-14
download: Fix dead GnuPG links

treib-holdings:
Two links in the download page's verification instructions return 404:

- `https://www.gnupg.org/download/index.en.html#binary`, the GnuPG download options link, used in the Windows and Linux sections
- `https://www.gnupg.org/gph/en/manual/x334.html`, the "GNU handbook section on key management" link, used in the Windows and macOS sections

gnupg.org dropped the `index.en.html` variant and consolidated the GNU Privacy Handbook into a single page. This points them at `https://www.gnupg.org/download/index.html#binary` and `https://www.gnupg.org/gph/en/manual.html#MANAGEMENT`, the Key Management chapter anchor in the consolidated handbook. Both return 200 and both anchors exist on the target pages.

The two URLs are assigned once at the top of `_includes/templates/download.html`, so all four rendered links on https://bitcoincore.org/en/download/ are fixed by this change.


fanquake:
ACK 5da66c7d4a542297e43f06899ce92665bd5ebf79

